PR #25, opened 2023-09-13
Feature/menu bar

sanaf1333:


evshi:
@sanaf1333 when merging this branch I got a ton of eslint errors

packages/runtime/src/widgets/DataGrid.tsx
```diff
@@ -0,0 +1,115 @@
+import React from "react";
```
evshi, line 1: Please remove this file

evshi:
A few small things I noticed during testing:

- Side menu height is not responsive
- Need support to specify different icons
- Collapse animation is a little janky because the text sizing isn't smooth

Also, let's move this up to the View widget property like we discussed and render this from the EnsembleScreen as part of the frame.

sanaf1333:
> @sanaf1333 when merging this branch I got a ton of eslint errors

can you tell me what eslint errors you are getting?

evshi:
Closing since this was merged with #35

PR #28, opened 2023-09-13
Charts widget added

evshi:


evshi:
Couple things missing from this @tibrahul:

- Data should be bound dynamically
- Need to be able to control chart sizing

evshi:
Closing since this was merged with #35

PR #29, opened 2023-09-14
Add a few widgets

vusters:
- add Image/Icon widgets
- add mainAxis, crossAxis to Row/Column
- Card is ongoing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,5 @@ yarn-error.log*
 
 # vercel
 .vercel
+.idea
+pnpm-lock.yaml
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -22,5 +22,9 @@
     "react": "^18.2.0",
     "tsconfig": "workspace:*",
     "tsup": "^7.2.0"
+  },
+  "dependencies": {
+    "@mui/icons-material": "^5.14.9",
+    "@mui/material": "^5.14.9"
   }
 }
--- a/packages/runtime/src/util/types.ts
+++ b/packages/runtime/src/util/types.ts
@@ -1,0 +1,19 @@
+import type { Widget } from "framework";
+
+export interface EnsembleWidgetProps {
+  id?: string;
+  [key: string]: unknown;
+}
+
+export type FlexboxProps = {
+  children: Widget[];
+  mainAxis?: string;
+  crossAxis?: string;
+  gap?: number;
+} & EnsembleWidgetProps;
+
+export type IconProps = {
+  name: string;
+  size?: number;
+  color?: string;
+} & EnsembleWidgetProps;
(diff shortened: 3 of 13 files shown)


.gitignore
```diff
@@ -35,3 +35,5 @@ yarn-error.log*
 
 # vercel
 .vercel
+.idea
+pnpm-lock.yaml
```
evshi, line 39: pnpm-lock file should be checked in

packages/runtime/package.json
```diff
@@ -22,5 +22,9 @@
     "react": "^18.2.0",
     "tsconfig": "workspace:*",
     "tsup": "^7.2.0"
+  },
+  "dependencies": {
+    "@mui/icons-material": "^5.14.9",
+    "@mui/material": "^5.14.9"
```
evshi, line 28: Please move this to dev dependencies. Since this is a package, transitive dependencies are usually not bundled and added to peer dependencies so it's only need during dev time.

packages/runtime/src/util/types.ts
```diff
@@ -0,0 +1,19 @@
+import type { Widget } from "framework";
+
+export interface EnsembleWidgetProps {
+  id?: string;
+  [key: string]: unknown;
+}
+
+export type FlexboxProps = {
```
evshi, line 8: Props should be colocated with the implementation

packages/runtime/src/widgets/Image.tsx
```diff
@@ -0,0 +1,38 @@
...
+export type ImageProps = {
+  source: string;
+
+  // move these under styles
+  width?: number | string;
+  height?: number | string;
+  fit?: "contain" | "cover" | "fill" | "none" | "scale-down";
+  borderRadius?: number;
+  borderWidth?: number;
+  borderColor?: number | string;
```
evshi, line 15: I think these are under a separate styles object?

evshi (commented):
Mostly looks good, just need to clean up some lines but I can do that after

packages/runtime/src/util/types.ts
vusters: This is being used by Row/Column/Flex hence it's outside.

packages/runtime/src/widgets/Image.tsx
vusters: yea, I will introduce a styles object.

packages/runtime/src/util/types.ts
evshi: Got it, I'll make a Layout folder and put colocate all the containers and their typings in there.

evshi (approved):
Go ahead and merge, will clean up after

evshi:
Closing since this was merged in #35

PR #63, opened 2023-09-24
Widget/avatar

sanaf1333:


packages/runtime/src/widgets/Avatar.tsx
```diff
@@ -0,0 +1,137 @@
...
+    height?: number | string;
+    backgroundColor?: string;
+  };
+  menu?: AvatarMenu[];
+};
+
+export const Avatar: React.FC<AvatarProps> = (props) => {
+  const [code, setCode] = useState("");
+  const [screen, setScreen] = useState("");
+  function stringToColor(string: Expression<string>) {
```
evshi, line 40: Move this outside the component

packages/runtime/src/widgets/Avatar.tsx
```diff
@@ -0,0 +1,137 @@
...
+    menuItem.onTap?.executeCode && setCode(menuItem.onTap?.executeCode);
+    menuItem.onTap?.navigateScreen && setScreen(menuItem.onTap?.navigateScreen);
+    handleMenuClose();
+  };
+  useEffect(() => {
+    code && onTapCallback();
+  }, [code]);
+  useEffect(() => {
+    screen && onNavigate();
+  }, [screen]);
```
evshi, line 93: Same as above use in click handler instead of using extra variables.

packages/runtime/src/widgets/Avatar.tsx
```diff
@@ -0,0 +1,137 @@
...
+  };
+
+  const handleMenuClick = (menuItem: any) => {
+    menuItem.onTap?.executeCode && setCode(menuItem.onTap?.executeCode);
+    menuItem.onTap?.navigateScreen && setScreen(menuItem.onTap?.navigateScreen);
+    handleMenuClose();
+  };
+  useEffect(() => {
+    code && onTapCallback();
+  }, [code]);
```
evshi, line 90: Just move this into the menu click handler. No need for extra state variables.

packages/runtime/src/util/renderMuiIcons.tsx
```diff
@@ -0,0 +1,20 @@
+import * as MuiIcons from "@mui/icons-material";
+
+export const renderMuiIcon = (
```
evshi, line 3: Make this an Icon component instead of a util function

packages/runtime/src/widgets/Avatar.tsx
```diff
@@ -0,0 +1,137 @@
...
+    if (words.length === 1) {
+      return words[0][0].toUpperCase();
+    } else if (words.length >= 2) {
+      return `${words[0][0].toUpperCase()}${words[1][0].toUpperCase()}`;
+    }
+
+    return "";
+  };
+  const nameString = props.name?.toString();
+  const [menuAnchorEl, setMenuAnchorEl] = useState<null | HTMLElement>(null);
```
evshi, line 73: This should be a `ref`?

packages/runtime/src/widgets/Avatar.tsx
```diff
@@ -0,0 +1,137 @@
...
+        onClick={handleMenuOpen}
+      >
+        {props?.name
+          ? generateInitials(props.name)
+          : renderMuiIcon(props?.icon ?? "Person")}
+      </MuiAvatar>
+      {props.menu && (
+        <Menu
+          anchorEl={menuAnchorEl}
+          open={Boolean(menuAnchorEl)}
```
evshi, line 120: Use a separate boolean state variable for tracking menu open state not an element.

packages/runtime/src/widgets/Text.tsx
```diff
@@ -11,9 +11,6 @@ export type TextProps = {
 
 export const Text: React.FC<TextProps> = (props) => {
   const [text, setText] = useState(props.text);
-  if (props.text !== text) {
-    setText(props.text);
-  }
```
evshi, line 16: Not sure why you're removing this?

packages/runtime/src/widgets/Avatar.tsx
sanaf1333: we can't use the hook in clickHandler and since we have to pass dynamic values to the hook, I am setting up the variables, so we can execute the callback function returned from the hook for required value. If I place onNavigate() in the clickHandler it won't work due to the asynchronous nature (the state is set after the onNavigate) is executed. I'll try to find a work around to improve this.

packages/runtime/src/widgets/Avatar.tsx
sanaf1333: We can't call hook in click handler and the value passed to the hook is dynamic so I've setup a state for that. If I call onTapCallBack() in click handler, we'll get issues since the state value might be setting up after the onTapCallBack();

packages/runtime/src/widgets/Avatar.tsx
sanaf1333: its being used like this in documentation

packages/runtime/src/util/renderMuiIcons.tsx
sanaf1333: @evshi do you mean Icon widget?

packages/runtime/src/widgets/Text.tsx
sanaf1333: Peter Parker text doesn't change to Spiderman in presence of this condition

packages/runtime/src/util/renderMuiIcons.tsx
evshi: Hmm, actually there's already a util for this see `Icon.tsx`

packages/runtime/src/widgets/Text.tsx
evshi: Keep it as it was before for now, I will update this

packages/runtime/src/widgets/Avatar.tsx
evshi: The hook returns a callback, that callback can be called inside the click handler. The values in the callback will be scoped to whatever was passed into the hook.

packages/runtime/src/widgets/Text.tsx
sanaf1333: alright

evshi:
@sanaf1333 can you rebase this onto main to resolve the conflits?

sanaf1333:
> @sanaf1333 can you rebase this onto main to resolve the conflits?

sure, on it.

sanaf1333:
> > @sanaf1333 can you rebase this onto main to resolve the conflits?
> 
> sure, on it.

@evshi done

packages/runtime/src/widgets/Text.tsx
```diff
@@ -14,6 +14,9 @@ export const Text: React.FC<TextProps> = (props) => {
   const { values } = useRegisterBindings({ ...props, text }, props.id, {
     setText,
   });
+  if (props.text !== text) {
+    setText(props.text);
+  }
```
evshi, line 19: @sanaf1333 this can be removed now

packages/runtime/src/util/renderMuiIcons.tsx
evshi: @sanaf1333 can you address this when you can a chance?

evshi (approved):
Looks good, can you please rebase onto main and add the Avatar widget to the page?

packages/runtime/src/util/renderMuiIcons.tsx
sanaf1333: @evshi  fixed this, now using current Icon component for this instead of a function.

evshi (approved):
@sanaf1333 can you add a usage of the Avatar widget to the YAML?

PR #68, opened 2023-09-26
Implement GridView

anserwaseem:


packages/runtime/src/widgets/GridView.tsx
```diff
@@ -0,0 +1,90 @@
...
+    : useEnsembleStore((state) => ({
+        templateData: get(state.screen, data as string) as object,
+      })).templateData;
+
+  const namedData = map(templateData, (value) => ({
+    [name]: value,
+  }));
+
+  const rows = [];
+  const colCount = styles?.horizontalTileCount ?? 4;
```
evshi, line 44: Create const for default column count.

evshi (approved):

Issue #54, opened 2023-09-20
Implement Conditional widget


kmahmood74:
Follow the yaml configuration that exists for the Conditional widget for Ensemble flutter today

PR #78, opened 2023-09-27
Implement Conditional

anserwaseem:
## Describe your changes
Added Conditional widget as per schema

https://github.com/EnsembleUI/ensemble-react/issues/54

- [x] I have added tests


packages/runtime/src/widgets/Conditional.tsx
```diff
@@ -0,0 +1,96 @@
...
+  return [true, ""];
+};
+
+const extractWidget = (condition: CondtionalElement) => {
+  for (const key in condition) {
+    if (key !== "if" && key !== "elseif" && key !== "else")
+      return {
+        name: key,
+        properties: condition[key as keyof typeof condition] as {},
+      } as Widget;
```
evshi, line 86: There's a util for this called `unwrapWidget` in the framework package parser you can use here.

packages/runtime/src/widgets/Conditional.tsx
```diff
@@ -0,0 +1,96 @@
...
+export type ConditionalProps = {
+  conditions: CondtionalElement[];
+};
+
+export const Conditional: React.FC<ConditionalProps> = (props) => {
+  const [isValid, errorMessage] = hasProperStructure(props.conditions);
+  if (!isValid) throw Error(errorMessage);
+
+  let element = head(
+    props.conditions.filter((condition) => {
```
evshi, line 23: You can use `some` to avoid iterating over entire list.

evshi (approved):
Some small comments but GEFN

packages/runtime/src/widgets/Conditional.tsx
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  const widget = extractWidget(element);
+
+  return <>{EnsembleRuntime.render([widget])}</>;
+};
+
+WidgetRegistry.register("Conditional", Conditional);
+
+const hasProperStructure = (
```
evshi, line 49: These util functions would be good candidates for unit tests

packages/runtime/src/widgets/Conditional.tsx
```diff
@@ -0,0 +1,96 @@
...
+    | { if: Expression<string>; elseif?: never; else?: never }
+    | { elseif: Expression<string>; if?: never; else?: never }
+    | { else: null; if?: never; elseif?: never }
+  );
+
+export type ConditionalProps = {
+  conditions: CondtionalElement[];
+};
+
+export const Conditional: React.FC<ConditionalProps> = (props) => {
```
evshi, line 18: Would be nice to have a test for this

packages/runtime/src/widgets/Conditional.tsx
evshi: Yeah you can use the `evaluate` util instead of the hook to get the value immediately

packages/runtime/src/widgets/__tests__/Conditional.test.tsx
```diff
@@ -0,0 +1,233 @@
...
+        {
+          elseif: "true",
+          Text: { name: "Text", properties: { text: "Widget B" } },
+        },
+      ],
+    };
+
+    expect(() =>
+      render(<Conditional conditions={conditionalProps.conditions} />)
+    ).not.toThrow();
```
evshi, line 45: @anserwaseem take a look at the EnsembleApp.test.tsx file. You can assert that the elements you expect are rendered.

packages/runtime/src/widgets/__tests__/__resources__/helloworld.yaml
```diff
@@ -0,0 +1,41 @@
+View:
```
evshi, line 1: Are you using this file?

evshi (commented):
Thanks for adding tests! Left some improvement comments

packages/runtime/src/widgets/__tests__/Conditional.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+  it("extracts the correct widget", () => {
+    const condition = {
+      if: "true",
+      Text: { name: "Text", properties: { text: "Widget A" } },
+    };
+
+    const widget = extractWidget(condition);
+    expect(widget.name).toBe("Text");
+  });
```
evshi, line 214: @anserwaseem you have a bug here, the returned widget is malformed (you're unwrapping the widget twice):

```
     {
        name: 'Text',
        properties: { name: 'Text', properties: { text: 'Widget C' } }
      }
```

evshi (approved):

Issue #39, opened 2023-09-16
Implement multi select widget


evshi:
ACs:

- Create input box that allows you to select from multiple options
- Render selected options
- Ability to remove selected option

![Image](https://user-images.githubusercontent.com/4132143/268418835-c4e7f1fb-6acd-4180-b4e9-62431b216c2d.png)

evshi:
@anserwaseem can you please help complete this ticket from the open PR?

PR #79, opened 2023-09-28
Added Multi-Select widget

tibrahulz:
## Describe your changes
- Added Multi Select widget as per requirement.

https://github.com/EnsembleUI/ensemble-react/issues/39

- [ ] I have added tests


packages/runtime/src/widgets/MultiSelect.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const handleChange = (value: string[]) => {
+    setSelectedValue(value)
+  };
+
+  return (
+    <SelectComponent
+      mode="multiple"
+      style={{ width: width || '100%' }}
+      placeholder={placeholder || 'Select'}
```
evshi, line 48: Need to resolve this binding.

packages/runtime/src/widgets/MultiSelect.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { Expression } from 'framework';
+import { EnsembleWidgetProps } from '../util/types';
+
+type SelectOption = {
+  label: Expression<string>;
+  value: Expression<string | number>;
+}
+export type MultiSelectProps = {
+  placeholder?: Expression<string>;
+  data: SelectOption[];
```
evshi, line 15: Data must be a binding/expression

evshi (changes requested):
You need to use the `useRegisterBindings` to binding the values to the ensemble state so they can be read later.

apps/starter/src/ensemble/collections.yaml
```diff
@@ -17,3 +17,19 @@ View:
                     navigateModalScreen:
                       name: Ask Embrace
                       maskClosable: false
+        - Row:
+            mainAxis: spaceBetween
+            crossAxis: center
+            children:
+              - Markdown:
+                  text: "### Groups"
+              - MultiSelect:
```
evshi, line 26: Just sanity checking here, there's no existing widget in Ensemble that has this multi select functionality right?

packages/runtime/src/widgets/MultiSelect.tsx
```diff
@@ -0,0 +1,142 @@
+import React, { useEffect, useMemo, useState } from "react";
+import { Expression, useTemplateData } from "framework";
+import { WidgetRegistry } from "../registry";
+import { PlusCircleOutlined } from "@ant-design/icons";
+import { Select as SelectComponent, Space } from "antd";
+import { get } from "lodash-es";
+import { SearchStyles } from "../util/types";
+import { getColor } from "../util/utils";
+
+interface EnsembleWidgetProps<T> {
```
evshi, line 10: Can you import this type from utils for now? Will refactor later to clean up

packages/runtime/src/widgets/MultiSelect.tsx
```diff
@@ -0,0 +1,142 @@
...
+  placeholder,
+  styles,
+}) => {
+  const [options, setOptions] = useState<SelectOption[]>([]);
+  const [newOption, setNewOption] = useState("");
+  const [selectedValues, setSelectedValues] = useState<string[] | undefined>(
+    defaultOptions?.map((item) => item.value?.toString())
+  );
+
+  const templateData = useMemo(() => useTemplateData(data), [data]);
```
evshi, line 43: Shouldn't need useMemo here, useTemplateData already memoizes the value

evshi (approved):
Go ahead and merge if needed to unblock @sibteali786. We'll need probably need to do another pass at this like Search

apps/starter/src/ensemble/collections.yaml
anserwaseem: yes, right

evshi (approved):

PR #81, opened 2023-09-28
Added date and Date Range Picker

tibrahulz:
## Describe your changes
Date Widget 

https://github.com/EnsembleUI/ensemble-react/issues/58
https://github.com/EnsembleUI/ensemble-react/issues/44

- [ ] I have added tests


packages/runtime/src/widgets/DateRange.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { DateProps } from "./Date";
+
+const { RangePicker } = DatePicker;
+
+export const DateRange: React.FC<DateProps> = ({
+  format = 'YYYY/MM/DD',
+  styles
+}) => {
+
+  return <RangePicker
```
evshi, line 13: Need to bind this to Ensemble state so the values can be read

packages/runtime/src/widgets/Date.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const Date: React.FC<DateProps> = ({
+  value,
+  minDate,
+  minDateFormat,
+  maxDate,
+  maxDateFormat,
+  format = 'YYYY/MM/DD',
+  styles
+}) => {
+  const disableDate = (current: dayjs.Dayjs) => {
```
evshi, line 31: Need to bind to ensemble state so values can be read.

evshi (changes requested):
Please bind the widgets to ensemble state with `useRegisterBindings` so their value can be read

evshi (commented):
Looks good but you need to fix your build issues

Issue #65, opened 2023-09-25
Add Card widget


evshi:
ACs:

- [ ] Add a new widget for Card styling
- [x] Use Card in GridView to demonstrate Embrace user management dashboard

![Image](https://github.com/EnsembleUI/ensemble-react/assets/4132143/7874cdc7-d30d-4a2a-b85a-250162bd7a39)

sibteali786:
We can also have media as an object with properties like 
```yaml
media:
    src
    height
    width
    placeholder  // if there is no image to show, just show a placeholder here 


```

evshi:
The Card widget should be much more flexible. Just allow it accept children so people can build their own layout inside. The main value is in the styling.

PR #84, opened 2023-09-28
added card widget

sibteali786:
## Describe your changes
Added Card Widget
### Screenshots [Optional]
image

#65 
- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Card.tsx
+++ b/packages/runtime/src/widgets/Card.tsx
@@ -1,4 +1,4 @@
-import type { Widget } from "framework";
+import type { EnsembleWidget } from "framework";
 import { useMemo } from "react";
 import { WidgetRegistry } from "../registry";
 import type { EnsembleWidgetProps } from "../util/types";
@@ -6,7 +6,7 @@ import { EnsembleRuntime } from "../runtime";
 
 export type CardProps = {
   [key: string]: unknown;
-  children: Widget[];
+  children: EnsembleWidget[];
   styles?: {
     gap?: number;
     borderColor?: string;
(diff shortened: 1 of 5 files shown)


apps/starter/src/ensemble/home.yaml
```diff
@@ -193,8 +193,43 @@ View:
...
+                          name: user
+                          template:
+                            Card:
+                              styles:
+                                display: flex
+                                flexDirection: column
+                                justifyContent: space-between
+                                alignItems: flex-start
+                                gap: 16
+                                padding: 24
```
evshi, line 212: Can we just a Column widget inside instead of requiring this styling?

sibteali786:
@evshi i have added a very basic card widget for now, its yaml is similar to what Sana did used for CardComponent except Image property. This is what we can achieve with it. Although i am not sure about current yaml as it is more relevant to web than flutter. What do you suggest  ? 

Like for instance -> instead of crossAxis we are using justifyItems here

packages/runtime/src/widgets/Card.tsx
```diff
@@ -1,37 +1,67 @@
...
+      | "space-between"; //justify options
+    gap?: number;
+    border: string;
+    borderRadius: string;
+    borderWidth: string;
+    boxShadow: string;
+    width: number;
+    height: number;
+    margin: string;
+    padding: string;
```
evshi, line 38: See my comment above but let's not expose all these properties on the card widget. The purpose of the card widget is to provide the card styling, not to control the internal layout. Users can use other widgets like Row and Column to control internal layout.

Some styles that we _can_ expose for Card would be for:

shadow
border
border-radius
padding

sibteali786: Okay, I get it.  One question though, let us suppose if a user wants more gap in between items of card ? should we allow it or not. How can we control it if we want to for some reason ?

sibteali786: Okay i get it, let me update it

evshi (approved):
Looks good! Please rebase the latest changes for the home yaml

PR #90, opened 2023-09-29
Widget/tab bar

sanaf1333:

- [ ] I have added tests


packages/framework/src/parser.ts
```diff
@@ -64,6 +66,17 @@ export const unwrapWidget = (obj: Record<string, unknown>): Widget => {
...
+  if (isArray(items)) {
+    const valueItems =(items as Array<any>).map(({ label, widget, icon }) => {
+      const unwrappedWidget = unwrapWidget(widget);
+      return { label, icon, widget: unwrappedWidget }
+    });
+    set(properties as Object, "items", valueItems);
+  }
+  if (isObject(widget)) {
+    const unwrappedWidget = unwrapWidget(widget);
+    set(properties as object, "widget", unwrappedWidget);
```
evshi, line 78: This shouldn't be needed any more because it's taken care of by the block above

packages/runtime/src/runtime/runtime.tsx
```diff
@@ -53,6 +53,7 @@ export const EnsembleRuntime = {
     return <WidgetFn {...rootWidget.properties} />;
   },
   render: (widgets: Widget[]): ReactNode[] => {
+    //console.log("runtimeeeeeeeee", widgets);
```
evshi, line 56: Remove console

packages/runtime/src/widgets/Icon.tsx
```diff
@@ -4,7 +4,9 @@ import type { IconProps } from "../util/types";
 import { getColor, getIcon } from "../util/utils";
 
 export const Icon: React.FC<IconProps> = (props) => {
+  
   const IconComponent = getIcon(props.name);
+  //console.log(props, IconComponent);
```
evshi, line 9: Remove console

packages/runtime/src/widgets/SideBar.tsx
```diff
@@ -35,7 +35,7 @@ interface MenuItem {
 }
 
 interface MenuBaseProps {
-  items: MenuItem[];
+  menuItems: MenuItem[];
```
evshi, line 38: Is this because of the parser changes?

packages/runtime/src/widgets/TabBar.tsx
```diff
@@ -0,0 +1,107 @@
+import React from "react";
+import { WidgetRegistry } from "../registry";
+import { EnsembleRuntime } from "../runtime";
+import { Tabs, ConfigProvider } from "antd";
+import { Widget } from "framework";
+import { IconProps } from "../util/types";
+import { Icon } from "./Icon";
+const { TabPane } = Tabs;
+
+export type items = {
```
evshi, line 10: Rename `TabBarItem`

packages/runtime/src/widgets/TabBar.tsx
```diff
@@ -0,0 +1,107 @@
...
+  label: string;
+  icon?: IconProps;
+  widget: Widget;
+};
+export type TabBarProps = {
+  id?: string;
+  selectedIndex?: string;
```
evshi, line 17: should be number

packages/runtime/src/widgets/TabBar.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { EnsembleRuntime } from "../runtime";
+import { Tabs, ConfigProvider } from "antd";
+import { Widget } from "framework";
+import { IconProps } from "../util/types";
+import { Icon } from "./Icon";
+const { TabPane } = Tabs;
+
+export type items = {
+  label: string;
+  icon?: IconProps;
```
evshi, line 12: I don't think icon is part of the schema?

packages/runtime/src/widgets/TabBar.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    .ant-tabs > .ant-tabs-nav {
+      background-color: ${props.styles?.tabBackgroundColor || "white"};
+    }
+
+    .ant-tabs {
+      font-weight: ${props.styles?.tabFontWeight || "normal"};
+    }
+  `;
```
evshi, line 77: Is it possible to use design tokens here instead? https://ant.design/components/tabs#design-token

packages/runtime/src/widgets/SideBar.tsx
sanaf1333: yes, since menu items doesn't contain any widgets, it breaks.

packages/runtime/src/widgets/TabBar.tsx
sanaf1333: schema has icon: hasIcon
"properties": {
              "label": {
                "type": "string",
                "description": "Setting the tab label"
              },
              "icon": {
                "$ref": "#/$defs/HasIcon"
              },
              "widget": {
                "$ref": "#/$defs/Widget",
                "description": "Return an inline widget or specify a custom widget to be rendered as this tab's content"
              }
            }

packages/runtime/src/widgets/TabBar.tsx
sanaf1333: some of the style properties are not available in the design tokens, they can only be changed by overriding antd styles. I have added design token for those that are available. Also, setting global tokens might affect styling in children so I used all the possible component tokens.

packages/runtime/src/widgets/TabBar.tsx
evshi: You're right! My mistake

packages/runtime/src/widgets/SideBar.tsx
evshi: Can you check if this is still an issue after rebasing? I separated menu out from widgets and it's on another parsing path now.

If it's still an issue, let's just move the widget unwrapping into the TabBar component instead of the parser level.

packages/runtime/src/widgets/TabBar.tsx
evshi: Got it, let's keep an eye on this then because this might create issues with multiple tab bars.

packages/runtime/src/widgets/TabBar.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { WidgetRegistry } from "../registry";
+import { EnsembleRuntime } from "../runtime";
+import { Tabs, ConfigProvider } from "antd";
+import { Widget } from "framework";
+import { IconProps } from "../util/types";
+import { Icon } from "./Icon";
+const { TabPane } = Tabs;
+
+export type items = {
+  label: string;
```
evshi, line 11: These labels will need to be expressions also

packages/runtime/src/widgets/TabBar.tsx
sanaf1333: okay

packages/runtime/src/widgets/SideBar.tsx
sanaf1333: not an issue anymore with menu moved out from the widgets

evshi (approved):

PR #92, opened 2023-10-02
Implement navigateModalScreen

anserwaseem:

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/navigateModal.tsx
+++ b/packages/runtime/src/runtime/navigateModal.tsx
@@ -1,0 +1,55 @@
+import { useApplicationContext } from "framework";
+import { EnsembleScreen } from "./screen";
+import { Modal } from "antd";
+import { useState } from "react";
+import { EnsembleRuntime } from "./runtime";
+import { navigateModalScreenProps } from "../util/types";
+
+const navigateModalScreen = (
+  isModalVisible: boolean,
+  onCancel: () => void,
+  props?: string | navigateModalScreenProps
+) => {
+  const app = useApplicationContext();
+
+  const screenName = typeof props === "string" ? props : props?.name;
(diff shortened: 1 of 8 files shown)


packages/runtime/src/runtime/navigateModal.tsx
```diff
@@ -0,0 +1,55 @@
...
+) => {
+  const app = useApplicationContext();
+
+  const screenName = typeof props === "string" ? props : props?.name;
+  const maskClosable = typeof props === "string" ? true : props?.maskClosable;
+
+  const screen = app?.application?.screens.find(
+    (s) => s.name.toLowerCase() === screenName?.toLowerCase()
+  );
+  const title = screen?.header && EnsembleRuntime.render([screen.header]);
```
evshi, line 21: Can use useMemo here.

packages/runtime/src/runtime/navigateModal.tsx
```diff
@@ -0,0 +1,55 @@
+import { useApplicationContext } from "framework";
+import { EnsembleScreen } from "./screen";
+import { Modal } from "antd";
+import { useState } from "react";
+import { EnsembleRuntime } from "./runtime";
+import { navigateModalScreenProps } from "../util/types";
+
+const navigateModalScreen = (
```
evshi, line 8: Any reason to separate this function? If not, can we just keep it all in one?

packages/runtime/src/util/types.ts
```diff
@@ -6,6 +6,11 @@ export interface EnsembleWidgetProps {
   styles?: Record<string, string | number>;
 }
 
+export type navigateModalScreenProps = {
```
evshi, line 9: Move this to `actions.ts` file and fix casing

evshi (approved):

Issue #67, opened 2023-09-25
Implement PopupMenu


evshi:
ACs:

- Render a view that will trigger a menu
- Allow specifying actions

![Image](https://github.com/EnsembleUI/ensemble-react/assets/4132143/32bb7b7d-fbae-4171-888d-3eb002aeb4e3)

sibteali786:
hey @evshi some questions
1. Do i have to implement a modal ? That can be opened using maybe a button like in ensemble as [here](https://studio.ensembleui.com/app/e24402cb-75e2-404c-866c-29e6c3dd7992/screen/92a5b5f5-7278-404c-b2a4-9e6e27a57aab)
Please elaborate it a bit more if possible, thanks.

evshi:
@sibteali786 there's an existing widget that defines the schema https://github.com/EnsembleUI/ensemble/blob/a86a8081cabdc76425a5652cb4bb726c1e74ac56/assets/schema/ensemble_schema.json#L852

1. You will have to implement a menu view, not a full screen modal. Please see the screenshot.
2. Yes you need to allow any type of ensemble action, according to the schema.

sanaf1333:
@anserwaseem https://github.com/EnsembleUI/ensemble-react/blob/demo/embrace/packages/runtime/src/widgets/Dropdown.tsx

PR #95, opened 2023-10-03
Popup menu

sibteali786:
## Describe your changes
Added a PopMenu Widget
#67 

- [ ] I have added tests


packages/runtime/src/widgets/PopupMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const { values } = useRegisterBindings(props, props.id);
+  const onTapCallback = useExecuteCode(onItemSelect, values);
+  const onNavigate = useNavigateScreen(props.onItemSelect?.navigateScreen);
+  const [widgetProps, setWidgetProps] = useState<EnsembleWidget>();
+  const menuItems = (
+    <Menu>
+      {props.items.map((item) => (
+        <Menu.Item
+          key={item.value}
+          onClick={onItemSelect ? onTapCallback : onNavigate}
```
evshi, line 56: You will need to pass in the value to the callback so you know which item was selected

packages/runtime/src/widgets/PopupMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+          ...mergedStyles,
+        }}
+      />
+    </AntdDropdown>
+  );
+};
+
+WidgetRegistry.register("PopupMenu", PopupMenu);
+
+function renderWidget(widget: EnsembleWidget): ReactNode {
```
evshi, line 100: You can just use EnsembleRuntime.render?

packages/runtime/src/widgets/PopupMenu.tsx
evshi: I'll have an update to the actions hook that will make this easier.

packages/runtime/src/widgets/PopupMenu.tsx
sibteali786: sure, will refactor that

evshi:
Closing since this was merged with demo branch

PR #97, opened 2023-10-04
Widget/dropdown

sanaf1333:
## Describe your changes
added dropdown widget

Closes #

- [ ] I have added tests


packages/runtime/src/widgets/Dropdown.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const onNavigate = useNavigateScreen(props.onChange?.navigateScreen || "");
+
+  const handleChange = (event: SelectChangeEvent) => {
+    setDropdownValue(event.target.value);
+    props.onChange?.executeCode && onTapCallback();
+    props.onChange?.navigateScreen && onNavigate();
+  };
+
+  return (
+    <FormControl
```
evshi, line 67: What is `FormControl` doing here? According to the docs, it provides context but only for a few components? 

https://mui.com/material-ui/api/form-control/#import

packages/runtime/src/widgets/Dropdown.tsx
```diff
@@ -0,0 +1,112 @@
+import React from "react";
+import type { Expression } from "framework";
+import { useRegisterBindings, useExecuteCode } from "framework";
+import { useNavigateScreen } from "../runtime/navigate";
+import { WidgetRegistry } from "../registry";
+import { MenuItem, Select, FormControl } from "@mui/material";
+import { SelectChangeEvent } from "@mui/material/Select";
```
evshi, line 7: Any reason to go with material UI over antD here?

evshi (approved):
GEFN, let's revisit later

PR #116, opened 2023-10-09
Add onValueChange and onSelect in Search widget

### LINKED ISSUE #96: Enhance Search widget
## AC:
- add `onValueChange` property that handles the search via action (invokeAPI) 

### Proposed YAML:
```yaml
Search:
  onValueChange: getEmails
```

or

```yaml
Search:
  onValueChange: 
    invokeAPI: getData
      inputs:
        limit: 100
```

### Additional context:
Extending the original [ticket](https://github.com/EnsembleUI/ensemble-react/issues/23) for Search widget

anserwaseem:
## Describe your changes
Added `onValueChange` and `onSelect` properties for Search widget

Closes #96

- [ ] I have added tests


packages/runtime/src/widgets/Search.tsx
```diff
@@ -18,31 +24,46 @@ export type SearchProps = {
...
   data,
   searchKey,
+  onValueChange,
+  onSelect,
   styles,
 }) => {
   const [options, setOptions] = useState<SelectProps<object>["options"]>([]);
+  const [value, setValue] = useState("");
+  // eslint-disable-next-line no-empty-pattern
+  const {} = useRegisterBindings(
```
evshi, line 43: You can just call useRegisterBindings without assigning the result to anything to avoid the linter error

packages/runtime/src/widgets/Search.tsx
```diff
@@ -52,16 +73,36 @@ export const Search: React.FC<SearchProps> = ({
               }))
           : [],
       );
+    } else if (onValueChange) {
```
evshi, line 76: This should be an else if. The value change handler should fire independently of the other properties. We can't assuming it is only going to be invokeAPI either.

packages/runtime/src/widgets/Search.tsx
```diff
@@ -52,16 +73,36 @@ export const Search: React.FC<SearchProps> = ({
...
+        if (apiResponse) {
+          setOptions(
+            findValuesByKey<string>(apiResponse, searchKey ?? "")
+              .filter((item) =>
+                item.toLowerCase().includes(search.toLowerCase()),
+              )
+              .map((item) => ({
+                value: item,
+                label: item,
+              })),
```
evshi, line 89: This block should be in YAML instead of hardcoded to into the component. Expose setOptions as a setter.

packages/runtime/src/widgets/Search.tsx
```diff
@@ -18,51 +24,108 @@ export type SearchProps = {
...
+          setFilteredOptions(
+            findValuesByKey<string>(apiResponse, searchKey)
+              .filter((item) =>
+                item.toLowerCase().includes(search.toLowerCase()),
+              )
+              .map((item) => ({
+                value: item,
+                label: item,
+              })),
+          );
```
evshi, line 75: The idea here is if onValueChange invokesApi, the API will return the filtered values and you just provide the data binding, e.g.

```
Search:
  options: mySearchEndpoint.body.results
  onValueChange:
     invokeApi
        name: mySearchEndpoint
        inputs:
           searchKey: value
```

evshi: So we don't need to do this in react code just invoke the callback if it's there

anserwaseem: got it

evshi (commented):
Let's de-prioritize this for now and we'll come back to this when we get more specific designs. I think it will help to to understand how this will be implemented

evshi:
Closing in favor of #242

PR #123, opened 2023-10-12
implement global modal and closeAllDialogs

### LINKED ISSUE #114: Refactor navigateModalScreen to use global portal so you don't have to render element
Currently the navigateModalScreen returns the Modal element and requires the consuming component to render the Modal. This is unintuitive and can create issues with multiple modals.

ACs:
- [ ] Change navigateModalScreen hook to no longer turn Modal element
- [ ] Implement global Modal render, e.g. w/ React portal
- [ ] navigateModalScreen should re-use global Modal

anserwaseem:
## Describe your changes
- Configured global Modal
- Only one modal displayed at a time
- Added `closeAllDialogs` action

Closes #114 

- [ ] I have added tests


packages/runtime/src/widgets/TabBar.tsx
```diff
@@ -47,7 +47,7 @@ export const TabBar: React.FC<TabBarProps> = (props) => {
     if (containsExpression(label)) {
       const cleanedExpression = label.replace(/\${|}/g, "");
       labelEvaluated = evaluate(
-        context as ScreenContextDefinition,
+        context as ScreenContextDefinition as any,
```
evshi, line 50: Unneeded change?

packages/runtime/src/runtime/hooks/useNavigateModal.tsx
```diff
@@ -22,47 +21,34 @@ export type UseNavigateModalScreenResult =
...
-
-  const app = useApplicationContext();
-
   const screenName = typeof action === "string" ? action : action?.name;
-  const maskClosable =
-    typeof action === "string" ? true : Boolean(action?.maskClosable);
+  const app = useApplicationContext();
 
-  const { screen, title } = useMemo(() => {
+  const openModal = (): void => {
```
evshi, line 29: useCallback

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -134,17 +135,23 @@ export const useEnsembleAction = (
...
-    executeCode ||
-    navigateScreen ||
-    navigateModalScreen ||
-    showToast
+    (action.invokeApi && invokeApi) ||
+    (action.executeCode && executeCode) ||
+    (action.navigateScreen && navigateScreen) ||
+    (action.showToast && showToast) ||
+    (action.navigateModalScreen && navigateModalScreen) ||
+    (keys(action).find((key) => key === "closeAllDialogs") &&
```
evshi, line 151: Why do you need to match keys here?

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -134,17 +135,23 @@ export const useEnsembleAction = (
...
+      closeAllDialogs) || {
+      callback: () => {},
```
evshi, line 153: There's no need to return a default value?

packages/runtime/package.json
```diff
@@ -36,6 +36,7 @@
     "react": "^18.2.0",
     "react-chartjs-2": "^5.2.0",
     "react-dom": "^18.2.0",
+    "react-global-modal-plus": "^0.0.1",
```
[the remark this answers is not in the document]
anserwaseem: I tried for a bit, then I went on searching for open source solution, found this package react-global-modal but it had stacking of modals not the way we want (i.e., first hide opened modal and only then open new modal). So, I created this package. It solves our problem!

packages/runtime/src/widgets/TabBar.tsx
anserwaseem: compiler was not happy without it, will check again

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
anserwaseem: it was giving a weird behaviour for me. it was calling useNavigateModalScreen even though I requested showToast or closeAllDialogs, so I went with this approach :)

packages/runtime/package.json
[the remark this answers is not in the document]
anserwaseem: sorry for being hasty, you were right, we actually needed our own implementation!

packages/runtime/src/runtime/modal.tsx
```diff
@@ -0,0 +1,113 @@
...
+import type { NavigateModalScreenAction } from "framework";
+import { createContext, useState } from "react";
+import { createPortal } from "react-dom";
+import { Outlet } from "react-router-dom";
+import OpenInFullIcon from "@mui/icons-material/OpenInFull";
+import CloseFullscreenIcon from "@mui/icons-material/CloseFullscreen";
+
+type ModalProps = {
+  title?: string | React.ReactNode;
+} & Omit<Exclude<NavigateModalScreenAction, string>, "name">;
```
evshi, line 11: Just put the property here explicitly, there's no value to tie it to the action property since it's React code.

packages/runtime/src/runtime/modal.tsx
```diff
@@ -0,0 +1,113 @@
...
+  setVisible: (visible: boolean) => void;
+  content: React.ReactNode;
+  setContent: (content: React.ReactNode) => void;
+  options: ModalProps;
+  setOptions: (options: ModalProps) => void;
+}
+
+export const ModalContext = createContext<ModalContextProps>({
+  visible: false,
+  setVisible: () => {},
```
evshi, line 24: Can we consolidate this API? You can expose one function that calls the hooks underneath the hood to abstract away complexity

evshi (approved):
Just some small comments but otherwise LGTM

evshi (approved):

PR #130, opened 2023-10-22
add Modal styles

### LINKED ISSUE #126: Configure modal position/sizing

anserwaseem:

Closes #126 

- [ ] I have added tests


packages/runtime/src/runtime/hooks/useNavigateModal.tsx
```diff
@@ -18,7 +18,16 @@ export const useNavigateModalScreen: EnsembleActionHook<
...
+      : Boolean(action.maskClosable);
+  const position =
+    typeof action !== "string" ? action?.styles?.position : undefined;
+  const height =
+    typeof action !== "string" ? action?.styles?.height : undefined;
+  const width = typeof action !== "string" ? action?.styles?.width : undefined;
+  const margin =
+    typeof action !== "string" ? action?.styles?.margin : undefined;
+  const padding =
+    typeof action !== "string" ? action?.styles?.padding : undefined;
```
evshi, line 30: Can make this check once and assign all at once with destructuring

packages/runtime/src/runtime/modal.tsx
```diff
@@ -24,21 +31,61 @@ export const ModalWrapper: React.FC = () => {
...
+      setModalWidth(contentNode?.clientWidth || 0);
+      setModalHeight(contentNode?.clientHeight || 0);
+    }
+  }, [visible, isFullScreen, content]);
 
   const openModal = (
     content: React.ReactNode,
     newOptions: ModalProps,
   ): void => {
+    setKey((prevKey) => prevKey + 1);
```
evshi, line 50: What's the purpose of the key?

anserwaseem: when a new modal is opened from within an already opened one, Modal title was not updating, so putting a unique key was the solution. Although we use `Math.random()` as well, what do you say?

evshi: I see, so it's forcing a re-render here as I thought. I think this incrementing is fine

evshi (approved):

Issue #115, opened 2023-10-07
Hook Dropdown and MultiSelect widgets into Form context


evshi:
Dropdown and MuiltiSelect need to implement the Form context in order to have their values set in Form state

ACs:

- [ ] Implement form context in Dropdown and MultiSelect
- [ ] onSubmit should contain values for dropdown and multiselect

Issue #107, opened 2023-10-06
Widget state should be unregistered on unmount


evshi:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #133, opened 2023-10-23
Multiselect form

sibteali786:
## Describe your changes
Added Dropdown and hooked multi select and Dropdown to form context 
#115 
Closes #
#115 

- [ ] I have added tests


packages/runtime/src/widgets/Form/TextInput.tsx
```diff
@@ -15,7 +15,7 @@ export const TextInput: React.FC<TextInputProps> = (props) => {
   return (
     <AntForm.Item
       label={values.label}
-      name={values.label}
+      name={values.id}
```
evshi, line 18: This should use the provided ID if it exists (not the resolved one), and then fallback to the label

packages/runtime/src/widgets/MultiSelect.tsx
```diff
@@ -77,51 +78,60 @@ const MultiSelect: React.FC<MultiSelectProps> = (props) => {
...
-        borderColor: styles?.borderColor
-          ? getColor(styles.borderColor)
-          : undefined,
+        margin: "0px",
       }}
-      value={values.value}
-    />
+    >
+      <SelectComponent
+        onChange={handleChange}
```
evshi, line 92: What was the issue with the infinite rendering?

packages/runtime/src/widgets/index.ts
```diff
@@ -22,3 +22,4 @@ export * from "./Divider";
 export * from "./Tag";
 export * from "./Carousel";
 export * from "./PopupMenu";
+export * from "./DropDown";
```
evshi, line 25: nit: rename file Dropdown

packages/runtime/src/widgets/MultiSelect.tsx
sibteali786: I was setting a wrong dependency in onChange which was conflicting with what Form was handling, also the id of the two multi select at different pages ( yaml ) was same causing conflicting default states. Which causes infinite re-redners

packages/runtime/src/widgets/Form/TextInput.tsx
sibteali786: so are you suggesting to use label then @evshi

packages/runtime/src/widgets/MultiSelect.tsx
evshi: Got it, yeah the same ID across multiple pages is a known issue #107 I'm planning to address soon

evshi (commented):
LGTM, but you have some changes to modal and package.json that don't seem part of your PR. Can you rebase and only include your changes?

evshi (approved):

Issue #138, opened 2023-10-25
Random backgroundColor for Tag widget instead of a static default value


sibteali786:
**Is your feature request related to a problem? Please describe.**
While I was working with tags I felt a need to have a random background for different tags across my item-template array values 

**Describe the solution you'd like**
It would be nice to have random colors assigned to each tag instead of a static color, Though we can still opt for a static color. If the user provides a static color value it applies to that tag otherwise a random color is assigned to the tag.

**Additional context**
As in this screenshot we have random colors for each tag.
<img width="448" alt="Screenshot 2023-10-25 at 6 04 12 PM" src="https://github.com/EnsembleUI/ensemble-react/assets/62849614/dc2d5304-8c21-43ee-8afe-f3291a6295cd">

PR #139, opened 2023-10-25
Added random backgroundColor for tag if no default provided

sibteali786:
## Describe your changes
Just added a new way to have random tag backgroundColors 
#138 
Closes #

- [ ] I have added tests


evshi:
@sibteali786 I don't think this adds a lot of value for the customer, even though it's nice for development. Almost always the color will be tied to the label to indicate a specific meaning. I'm going to close this for now and we can revisit later if a need comes up.

Issue #41, opened 2023-09-16
Add chart title and labels


evshi:
ACs:

- Configure chart title from YAML
- Configure labels from YAML

PR #146, opened 2023-10-27
add js config property to bar chart

### LINKED ISSUE #71: Improve ChartJs data binding
Improve the data binding capabilities. Most importantly, the ChartJs config which specifies the data can be interpreted as JS, see flutter side

```dart
      'config': (value) {
        if (value is Map) {
          _controller.config = JSInterpreter.toJSString(value);
        } else {
          _controller.config = value;
        }
      }
```

ACs:
- [ ] update ChartJs component to evaluate config with screen context (if necessary)

anserwaseem:

Closes #71, #41, #72

- [ ] I have added tests


packages/runtime/src/widgets/Charts/BarChart.tsx
```diff
@@ -27,23 +33,42 @@ const options: ChartOptions<"bar"> = {
...
+      backgroundColor:
+        typeof dataset.backgroundColor === "string" &&
+        dataset.backgroundColor.startsWith("${")
+          ? (evaluate(
+              context as ScreenContextDefinition,
+              dataset.backgroundColor,
+            ) as string | string[])
+          : dataset.backgroundColor,
+    }));
+  }, [config?.datasets]);
```
evshi, line 66: You can pass in the entire string to the evaluate function. You may need to update the evaluate function's return logic to handle multiline object declarations.

packages/framework/src/evaluate.ts
```diff
@@ -63,8 +73,41 @@ export const evaluate = (
...
+          0,
+          open - (isArray(value) ? 1 : 0),
+        )}${replaced}${result.slice(close + 1 + (isArray(value) ? 1 : 0))}`;
+
+        searchIndex = open + toLength(value);
+      }
+      isString(result) && (result = JSON.parse(result));
+
+      return result;
+    }
```
evshi, line 110: I think this is overcomplicating things. If the value of js is a multiline string, then we can just evaluate the entire block, no need to evaluate the expressions individually.

The main problem we have to solve is whether you return an object or an interpolated string and expect it to get parsed later on. We might need an extra param for that since I'm not sure there's a good heuristic to determine what to return from the JS alone. Maybe just checking the object braces is enough....

packages/runtime/src/widgets/Charts/BarChart.tsx
```diff
@@ -27,23 +33,26 @@ const options: ChartOptions<"bar"> = {
...
+    () => evaluate(context as ScreenContextDefinition, config?.data?.datasets),
+    [config?.data?.datasets, context],
+  );
+
+  const { values } = useRegisterBindings({ labels, title }, id, {
+    setLabels,
+    setTitle,
+  });
+
+  console.log("BarChart", evaluatedDatasets);
```
evshi, line 53: Remove console.log

packages/runtime/src/widgets/Charts/BarChart.tsx
```diff
@@ -27,23 +33,26 @@ const options: ChartOptions<"bar"> = {
...
-  title?: Expression<string>;
-  [key: string]: unknown;
-} & EnsembleWidgetProps;
+  const [title, setTitle] = useState(config?.title);
+  const [labels, setLabels] = useState<string[]>(config?.data?.labels || []);
+  const context = useScreenContext();
 
-export const BarChart: React.FC<BarChartProps> = (props) => {
-  const { labels, datasets, title, styles } = props;
+  const evaluatedDatasets = useMemo(
```
evshi, line 43: Factor this out into the parent component?

packages/framework/src/evaluate.ts
```diff
@@ -1,10 +1,10 @@
-import { isEmpty, last, merge } from "lodash-es";
+import { isEmpty, isObjectLike, last, merge, toString } from "lodash-es";
 import type { InvokableMethods, ScreenContextDefinition } from "./state";
 import { EnsembleStorage } from "./storage";
 
 export const buildEvaluateFn = (
   screen: ScreenContextDefinition,
-  js?: string,
+  js?: unknown,
```
evshi, line 7: This should remain a string, any formatting should happen in the invoking component

packages/runtime/src/widgets/Charts/index.tsx
```diff
@@ -59,13 +65,19 @@ const tabsConfig = {
 };
 
 export const Chart: React.FC<ChartProps> = (props) => {
-  const { type } = props;
+  // eslint-disable-next-line @typescript-eslint/no-implied-eval, no-new-func
+  const config = new Function(
+    `return ${props.config?.toString() || "{}"}`,
+  )() as ChartConfigs;
```
evshi, line 71: You can just serialize the config directly no need to wrap as a separate function

evshi (changes requested):

packages/runtime/src/widgets/Charts/index.tsx
anserwaseem: Cannot use `JSON.parse` directly as it has some limitations such as https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/JSON_bad_parse#property_names_must_be_double-quoted_strings

packages/runtime/src/widgets/Charts/index.tsx
```diff
@@ -59,13 +70,24 @@ const tabsConfig = {
...
+        context as ScreenContextDefinition,
+        props?.config?.toString()?.replace(/['"]\$\{([^}]*)\}['"]/g, "$1"),
+      ) as ChartConfigs,
+    [props.config, context],
+  );
 
-  if (!type) {
-    return <b>Chart type missing</b>;
-  }
+  if (!config) return null;
```
evshi, line 84: We can return an Error widget or something to let them know the config is bad/missing. See the runtime render method for an example.

evshi (commented):
Mostly LGTM, just some small things

packages/runtime/src/widgets/Charts/index.tsx
```diff
@@ -59,13 +70,24 @@ const tabsConfig = {
...
 
 export const Chart: React.FC<ChartProps> = (props) => {
-  const { type } = props;
+  const context = useScreenContext();
+
+  const config = useMemo(
+    () =>
+      evaluate(
+        context as ScreenContextDefinition,
+        props?.config?.toString()?.replace(/['"]\$\{([^}]*)\}['"]/g, "$1"),
```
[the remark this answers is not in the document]
anserwaseem: yes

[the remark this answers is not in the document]
evshi: This just replaces all expression braces right? I wonder why not just have the config written as a javascript object to begin with?

evshi (approved):

Issue #32, opened 2023-09-14
Support labels on horizontal stacked bar chart


evshi:
![Image](https://github.com/EnsembleUI/ensemble-react/assets/4132143/60de65ad-4ee8-4859-9e0f-d18b4b265fbb)

PR #147, opened 2023-10-27
Added labels in horizontal stacked bar chart

Alwaz:
Closes https://github.com/EnsembleUI/ensemble-react/issues/32

- [ ] I have added tests


evshi (approved):
What was the issue here?

Alwaz:
> What was the issue here?

The labels were added in the YAML but the option to display those labels was set to `false`. I set that to true, displayed them on top and adjusted the height accordingly to make them visible.

Issue #148, opened 2023-10-27
Add workflow progress (aka stepper) widget


evshi:
**Is your feature request related to a problem? Please describe.**
Multistep workflows are common in applications, and web applications often use the screen real estate to visualize workflow progress. Configuring this workflow progress view is complicated without a dedicated widget.

**Describe the solution you'd like**

- [x] Define YAML properties for a workflow progress widget
- [x] Implement YAML with a react component, e.g. MUI react-stepper 

**Describe alternatives you've considered**

**Additional context**

PR #151, opened 2023-10-30
Workflow widget

sibteali786:
## Describe your changes
Added Workflow widget for the use case of showing step wise screens
#148 
Closes #
#148 

- [ ] I have added tests

### DIFF
--- a/apps/starter/src/ensemble/dashboard.yaml
+++ b/apps/starter/src/ensemble/dashboard.yaml
@@ -139,41 +139,86 @@ View:
             text: ${getDummyJson.body.total}
         - Stepper:
             id: stepper1
-            activeStepIndex: 0
             styles:
               connectorColor: "#31C48D"
               connectorHeight: 4
+            item-template:
+              data:
+                - name: Match
+                  iconName: ScaleOutlined
+                - name: SubmitDocs
+                  iconName: FindInPageOutlined
+                - name: Review
+                  iconName: Check
+              name: step
+              template:
+                Steps:
+                  children:
+                    # InActive
+                    - Column:
+                        crossAxis: center
+                        children:
+                          - Icon:
+                              name: ${step.iconName}
+                              color: grey
+                              styles:
+                                borderWidth: 1
+                                borderRadius: 30
+                                padding: 8
+                                backgroundColor: white
+                          - Text:
+                              text: ${step.name}
+                    # Active
+                    - Column:
+                        crossAxis: center
+                        children:
+                          - Icon:
+                              name: ${step.iconName}
+                              color: "#1D1C21"
+                              styles:
+                                borderWidth: 1
+                                borderRadius: 30
+                                padding: 8
+                                backgroundColor: white
+                          - Text:
+                              text: In Progress
+                    # Completed
+                    - Column:
+                        crossAxis: center
+                        children:
+                          - Icon:
+                              name: ${step.iconName}
+                              color: "#31C48D"
+                              styles:
+                                borderWidth: 1
+                                borderRadius: 30
+                                padding: 8
+                                backgroundColor: white
+                          - Column:
+                              gap: 4
+                              crossAxis: center
+                              children:
+                                - Text:
+                                    text: ${step.name}
+                                    styles:
+                                      fontSize: 14
+                                      fontWeight: 600
+                                - Row:
+                                    gap: 3
+                                    children:
+                                      - Text:
+                                          text: 12/31/3131
+                                          styles:
+                                            fontSize: 12
+                                            color: "#747184"
+                                            fontWeight: 500
+                                      - Icon:
+                                          name: Check
+                                          color: "#31C48D"
+                                          size: 16
+
             steps:
               - stepLabel: Match
-                inactiveStepWidget:
-                  Column:
-                    crossAxis: center
-                    children:
-                      - Icon:
-                          name: ScaleOutlined
-                          color: grey
-                          styles:
-                            borderWidth: 1
-                            borderRadius: 30
-                            padding: 8
-                            backgroundColor: white
-                      - Text:
-                          text: Match
-                activeStepWidget:
-                  Column:
-                    crossAxis: center
-                    children:
-                      - Icon:
-                          name: ScaleOutlined
-                          color: "#31C48D"
-                          styles:
-                            borderColor: "#31C48D"
-                            borderWidth: 1
-                            borderRadius: 30
-                            padding: 8
-                            backgroundColor: white
-                      - Text:
-                          text: Match
                 contentWidget:
                   Column:
                     crossAxis: center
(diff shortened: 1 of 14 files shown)


apps/starter/src/ensemble/dashboard.yaml
```diff
@@ -264,14 +264,145 @@ View:
...
         - Text:
             text: ${getDummyJson.body.total}
+        - Stepper:
+            id: stepper1
+            styles:
+              connectorColor: "#31C48D"
+              connectorHeight: 2
+            steps:
+              - stepLabel: Match
+                inactiveStepWidget:
```
sibteali786, line 278: @evshi this widget has three states active, inactive and completed. We can also have error state ( not implemented yet ). What are your thoughts for given Yaml syntax ?

sibteali786:
also @evshi , I just went on to see a new error , though I have registered the `activeStep` state in the useRegister bindings but if I am using `next` and `previous` buttons from the yaml for first time when the Widget loads they do not work ? while after I have used the buttons ( inside component for testing ) then I can simply use these yaml button to manipulate the `activeStep` ? any idea

apps/starter/src/ensemble/dashboard.yaml
evshi: I think you will want a property that allows you to define the workflow definition like charts does with data. Then you can provide an item-template that will bind the view to each step definition.

packages/framework/src/parser.ts
```diff
@@ -149,14 +150,43 @@ export const unwrapWidget = (obj: Record<string, unknown>): EnsembleWidget => {
...
+            widget as Record<string, unknown>
           );
           return { label, icon, widget: unwrappedWidget };
-        },
+        }
       );
       set(properties as object, "items", valueItems);
     }
   }
+  if (isArray(steps) && !isEmpty(steps)) {
```
evshi, line 161: Let's move this into your stepper widget since it's pretty specific

packages/tsconfig/base.json
```diff
@@ -1,21 +1,23 @@
...
-    "noUnusedLocals": false,
-    "noUnusedParameters": false,
-    "preserveWatchOutput": true,
-    "skipLibCheck": true,
-    "strict": true,
-    "strictNullChecks": true
-  },
-  "exclude": ["node_modules"]
-}
+	"$schema": "https://json.schemastore.org/tsconfig",
```
evshi, line 2: What are these changes for?

packages/runtime/src/widgets/Stepper.tsx
```diff
@@ -0,0 +1,152 @@
...
+        alternativeLabel
+        connector={
+          <CustomConnector
+            connectorColor={props?.styles.connectorColor}
+            connectorHeight={props?.styles.connectorHeight}
+          />
+        }
+      >
+        {values?.steps[0].inactiveStepWidget &&
+        values?.steps[0].activeStepWidget
```
evshi, line 80: Yeah this should be refactored in an item template and rendered like a DataGrid's Cell component

evshi:
@sibteali786 if you're having trouble with the dynamic data binding, let's close it out for now as is and make a separate ticket to refactor this later on.

sibteali786:
@evshi agree, let me fix the mentioned issues and then we can close it for now. In the meantime I can try to fix the data binding issue on weekend.

packages/tsconfig/base.json
sibteali786: these are prettier changes ig, let me undo it

sibteali786:
@evshi for now i haven't removed logic for stepper from parser. Would refactor it in another ticket for it

packages/runtime/src/widgets/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+const Stepper: React.FC<StepperProps> = (props) => {
+  const [activeStep, setActiveStep] = useState(0);
+  const itemTemplate = props["item-template"];
+  const templateData = useTemplateData(itemTemplate.data);
+  const namedData = map(templateData, (value) => {
+    const namedObj: Record<string, unknown> = {};
+    namedObj[itemTemplate.name] = value;
+    return namedObj;
+  });
```
evshi, line 57: This should be memoized

apps/starter/src/ensemble/dashboard.yaml
```diff
@@ -265,14 +265,149 @@ View:
...
+                  iconName: FindInPageOutlined
+                - name: Review
+                  iconName: Check
+                - name: Enroll
+                  iconName: Search
+              name: step
+              template:
+                Conditional:
+                  conditions:
+                    - if: ${index < stepsLength - 1}
```
evshi, line 322: Why is this check needed?

apps/starter/src/ensemble/dashboard.yaml
```diff
@@ -265,14 +265,149 @@ View:
...
+                      Column:
+                        crossAxis: center
+                        children:
+                          - Icon:
+                              name: ${step.iconName}
+                              color: ${!active ? completed ? "#31C48D" :"grey" :"#1D1C21"}
```
evshi, line 328: Should be `step.active`?

packages/framework/src/parser.ts
evshi: Please address this @sibteali786

packages/runtime/src/widgets/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,167 @@
...
+        alternativeLabel
+        connector={
+          <CustomConnector
+            connector_color={props?.styles.connectorColor}
+            connector_height={props?.styles.connectorHeight}
+          />
+        }
+      >
+        {values?.["item-template"]
+          ? namedData?.map((data, index) => (
```
evshi, line 91: We can return early if the props are not behaving as expected. We can render something like the `UnknownWidget` but maybe call it like `WarningWidget` or something with the issue. Gives good user feedback and keeps code cleaner.

packages/runtime/src/widgets/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,167 @@
...
+    },
+    [`& .${stepConnectorClasses.line}`]: {
+      borderColor: "#eaeaf0",
+      borderTopWidth: props?.connector_height ?? 3,
+      borderRadius: 1,
+    },
+  })
+);
+
+const CustomStepIcon = (
```
evshi, line 153: What is the difference between this and `StepType` component?

evshi (commented):
Heading in the right direction! Mainly just needs some cleanup

apps/starter/src/ensemble/dashboard.yaml
sibteali786: this check is for the final step, this yaml can be improved but for now it checks to make sure final step goes to else condition so that its shown in only two states inactive, completed.

apps/starter/src/ensemble/dashboard.yaml
sibteali786: the active, completed states are exposed from within component, so yeah can me made as step.active but is it necessary @evshi .

packages/runtime/src/widgets/Stepper/Stepper.tsx
sibteali786: This can be called as a wrapper only where I am destructuring some data,  I think I can simplify it further

packages/runtime/src/widgets/Stepper/Stepper.tsx
sibteali786: Good idea would be much better than returning null otherwise

apps/starter/src/ensemble/dashboard.yaml
evshi: Yes, it makes it consistent with the rest of the expressions, otherwise it's unclear where the active variable is coming from.

apps/starter/src/ensemble/dashboard.yaml
evshi: Okay this is fine for now then

evshi (approved):
I think this is a good first version! Let's continue to improve on it.

PR #152, opened 2023-10-31
add sticky Header and Footer in view

### LINKED ISSUE #137: Support sticky header and footer widget in View
ACs:

- [ ] Support header property of View as specified in schema
- [ ] Support footer property of View as specified in schema
- [ ] Header and footer should be fixed elements (they don't scroll)

anserwaseem:

Closes #137

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/parser.ts
+++ b/packages/framework/src/parser.ts
@@ -15,6 +15,8 @@ import type {
   EnsembleWidget,
   EnsembleAppModel,
   EnsembleMenuModel,
+  EnsembleHeaderModel,
+  EnsembleFooterModel,
 } from "./shared/models";
 import type { ApplicationDTO } from "./shared/dto";
 
@@ -67,18 +69,22 @@ export const EnsembleParser = {
   ): EnsembleScreenModel | EnsembleMenuModel => {
     const view = get(screen, "View");
     const viewNode = get(view, "body");
-    const header = get(view, "header") as Record<string, unknown> | undefined;
+    const header = get(view, "header") as EnsembleHeaderModel | undefined;
+    const footer = get(view, "footer") as EnsembleFooterModel | undefined;
 
     if (!viewNode) {
       throw new Error("Invalid screen: missing view widget");
     }
     const viewWidget = unwrapWidget(viewNode);
+    const body = unwrapBody(viewWidget, header, footer);
     const apis = unwrapApiModels(screen);
+
     return {
       ...(view ?? {}),
       name,
-      header: header ? unwrapWidget(header) : undefined,
-      body: viewWidget,
+      header: header ? unwrapHeader(header) : undefined,
+      footer: footer ? unwrapFooter(footer) : undefined,
+      body,
       apis,
     };
   },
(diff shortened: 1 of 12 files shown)


packages/runtime/src/runtime/screen.tsx
```diff
@@ -26,9 +29,24 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({ screen }) => {
...
+
   return (
     <ScreenContextProvider screen={screen}>
+      {HeaderFn instanceof Function && (
+        <HeaderFn {...(headerWidget as EnsembleWidget)?.properties} />
+      )}
       <WidgetFn {...rootWidget.properties} />
+      {FooterFn instanceof Function && (
+        <FooterFn {...(footerWidget as EnsembleWidget)?.properties} />
+      )}
```
evshi, line 49: We should create separate components for Header and Footer.

packages/framework/src/parser.ts
```diff
@@ -162,3 +168,160 @@ export const unwrapWidget = (obj: Record<string, unknown>): EnsembleWidget => {
...
+    marginTop,
+    marginBottom,
+  };
+
+  return {
+    name: "Column",
+    properties: {
+      children: [viewWidget],
+      styles: defaultStyles,
+    },
```
evshi, line 205: Styling defaults and UI logic should be part of the component, not in the parser.

packages/framework/src/parser.ts
```diff
@@ -67,18 +69,22 @@ export const EnsembleParser = {
   ): EnsembleScreenModel | EnsembleMenuModel => {
     const view = get(screen, "View");
     const viewNode = get(view, "body");
-    const header = get(view, "header") as Record<string, unknown> | undefined;
+    const header = get(view, "header") as EnsembleHeaderModel | undefined;
+    const footer = get(view, "footer") as EnsembleFooterModel | undefined;
```
evshi, line 73: The parser is meant to take raw YAML and assert that it matches the typed objects, not just cast them directly.

evshi (changes requested):
Need to make these changes:

1. Parser is just to assert data format and integrity. Input is unknown, output is EnsembleHeaderModel or EnsembleFooterModel
2. Create new components for EnsembleHeader and EnsembleFooter
3. Move default styling logic, component into Component implementation instead of parser

packages/runtime/src/runtime/body.tsx
```diff
@@ -0,0 +1,60 @@
...
+    ? "56px"
+    : typeof footer?.styles?.height === "number"
+    ? footer?.styles?.height + "px"
+    : footer?.styles?.height || "0px";
+
+  // default body styles
+  const defaultStyles = {
+    height: `calc(100vh - ${marginTop} - ${marginBottom})`,
+    overflow: "auto",
+  };
```
evshi, line 57: Avoid doing calculations like this. You can just use position with top and bottom values instead of calculating the height.

packages/runtime/src/runtime/body.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  set(body, ["properties", "styles"], defaultStyles);
```
evshi, line 59: Instead of passing this as props into the widget, let's just create a container around the widget with these rules

packages/runtime/src/runtime/footer.tsx
```diff
@@ -0,0 +1,53 @@
...
+    backgroundColor: footer?.styles?.backgroundColor || "white",
+    height: footer?.styles?.height || "56px",
+    top: `calc(100vh - ${
+      (typeof footer?.styles?.height === "number"
+        ? footer?.styles?.height + "px"
+        : footer?.styles?.height) || "56px"
+    })`,
+  };
+
+  return {
```
evshi, line 46: Just use the Column react component.

packages/runtime/src/runtime/footer.tsx
```diff
@@ -0,0 +1,53 @@
...
+    justifyContent: "center",
+    alignItems: "center",
+    width: footer?.styles?.width || "100%",
+    backgroundColor: footer?.styles?.backgroundColor || "white",
+    height: footer?.styles?.height || "56px",
+    top: `calc(100vh - ${
+      (typeof footer?.styles?.height === "number"
+        ? footer?.styles?.height + "px"
+        : footer?.styles?.height) || "56px"
+    })`,
```
evshi, line 43: Again, avoid calc. In this case, just use bottom: 0

packages/runtime/src/runtime/header.tsx
```diff
@@ -0,0 +1,67 @@
...
+    display: "flex",
+    justifyContent: "center",
+    alignItems: header?.styles?.centerTitle ? "center" : "flex-start",
+    backgroundColor: header?.styles?.backgroundColor || "white",
+    height: header?.styles?.titleBarHeight || 56,
+  };
+
+  if (typeof title === "string") {
+    return {
+      name: "Column",
```
evshi, line 45: Just use the react component.

packages/framework/src/shared/models.ts
```diff
@@ -53,3 +54,24 @@ export interface EnsembleWidget {
...
+    backgroundColor?: string;
+    centerTitle?: boolean;
+    titleBarHeight?: string | number;
+    color?: string;
+    [key: string]: unknown;
+  };
+}
+
+export interface EnsembleFooterModel {
+  children: EnsembleWidget[];
```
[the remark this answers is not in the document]
anserwaseem: i referred to the examples in kitchen sink https://studio.ensembleui.com/app/e24402cb-75e2-404c-866c-29e6c3dd7992/screen/7hhWzE4mZyiC8fhN1XiV & https://studio.ensembleui.com/app/e24402cb-75e2-404c-866c-29e6c3dd7992/screen/huq4Og8l2f0FAcJi85Kd

packages/runtime/src/runtime/body.tsx
anserwaseem: it didn't work out for me!

packages/framework/src/shared/models.ts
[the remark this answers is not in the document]
evshi: Okay, strange it's not consistent but we can go with this for now

packages/runtime/src/runtime/body.tsx
evshi: Okay GEFN

evshi (approved):
LGTM, just resolve conflicts and feel free to merge

PR #156, opened 2023-11-07
added expression type for backgroundColor

sibteali786:
## Describe your changes
Some stylistic changes for Tag

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Tag.tsx
+++ b/packages/runtime/src/widgets/Tag.tsx
@@ -1,5 +1,13 @@
-import type { Expression } from "@ensembleui/react-framework";
-import { useRegisterBindings } from "@ensembleui/react-framework";
+import type {
+  Expression,
+  ScreenContextDefinition,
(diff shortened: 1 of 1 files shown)


packages/runtime/src/widgets/Tag.tsx
```diff
@@ -10,23 +18,28 @@ export interface TagProps {
...
   const { values } = useRegisterBindings({ ...props, text }, props.id, {
     setText,
   });
+  const screen = useScreenContext();
+  const scope = useCustomScope();
+  const backgroundColor = evaluate(
+    screen as ScreenContextDefinition,
+    props?.styles?.backgroundColor,
+    scope
+  );
```
evshi, line 38: You can just pass the expression through useRegisterBindings, it does the same thing and memoizes the value for you. It's important so you don't evaluate on each render.

evshi (approved):

PR #167, opened 2023-11-10
Fix for Nav2-46 and 45 in Atlas-React

sibteali786:
## Describe your changes
1. Fixes the labels displaying on top of y-axis in stack bar chart.
2. Fixes the search bar back color.
3. Fixes the borderRadius for bars in Stack Bar chart.

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Charts/index.tsx
+++ b/packages/runtime/src/widgets/Charts/index.tsx
@@ -67,9 +67,7 @@ export const Chart: React.FC<ChartProps> = (props) => {
   }
 
   return (
-    <div ref={rootRef} style={{ height: "100%", width: "100%" }}>
-      {cloneElement(tabsConfig[type], { ...props })}
-    </div>
+    <div ref={rootRef}>{cloneElement(tabsConfig[type], { ...props })}</div>
   );
 };
 
--- a/packages/runtime/src/widgets/Charts/index.tsx
+++ b/packages/runtime/src/widgets/Charts/index.tsx
@@ -67,7 +67,9 @@ export const Chart: React.FC<ChartProps> = (props) => {
   }
 
   return (
-    <div ref={rootRef}>{cloneElement(tabsConfig[type], { ...props })}</div>
+    <div ref={rootRef} style={{ height: "100%", width: "100%" }}>
+      {cloneElement(tabsConfig[type], { ...props })}
+    </div>
   );
 };
 
--- a/packages/runtime/src/widgets/Search.tsx
+++ b/packages/runtime/src/widgets/Search.tsx
@@ -3,7 +3,7 @@ import { useTemplateData, type Expression } from "@ensembleui/react-framework";
 import type { SelectProps } from "antd";
 import { AutoComplete, Input } from "antd";
 import { SearchOutlined } from "@mui/icons-material";
-import { get, isObject } from "lodash-es";
+import { get, isObject, noop } from "lodash-es";
 import { WidgetRegistry } from "../registry";
 import type { EnsembleWidgetProps, HasBorder } from "../shared/types";
 import { getColor } from "../shared/styles";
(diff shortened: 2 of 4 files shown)


packages/runtime/src/widgets/Search.tsx
```diff
@@ -81,7 +80,6 @@ export const Search: React.FC<SearchProps> = ({
             ? getColor(styles.borderColor)
             : undefined,
           boxShadow: "none",
-          backgroundColor: styles?.backgroundColor,
```
evshi, line 84: Why remove this instead of specifying color in the YAML?

sibteali786: actually this applies color only to search input and not to whole search bar. Which is not desired. imo it should applied to whole search bar and not only specific part of it

packages/runtime/src/widgets/Charts/index.tsx
```diff
@@ -67,9 +67,7 @@ export const Chart: React.FC<ChartProps> = (props) => {
   }
 
   return (
-    <div ref={rootRef} style={{ height: "100%", width: "100%" }}>
```
evshi, line 70: What issue was this causing? The intention was that this should expand to the parent container and all Chart elements would size relative to this.

packages/runtime/src/widgets/Search.tsx
evshi: One option is to use the widget ID from the registered binding and assign it to the root element. Then you can use the id in your CSS selector since it's generated from JS

packages/runtime/src/widgets/Charts/index.tsx
sibteali786: it was causing a visual issue in one of atlas dashboard design requirements. I tried to fix it for that but wasn;t working unless I removed it. Let me see if I can accommodate it as well.

sibteali786: @evshi i rechecked and here is the problem. 
- We can either leave it as is. 
  - which takes all the space if its a child of any container like Row, the other child unless specified its width is squeezed. 
Screenshot 2023-11-14 at 8 42 59 PM

```yaml
                    - Column:
                        styles:
                          width: 60%
                        gap: 2
                        children:
                          - Text:
                              text: Current Month Activity
                              styles:
                                fontSize: 14
                                fontWeight: 400
                                color: '#747184'
                          - Text:
                              id: totalEarning
                              styles:
                                fontSize: 24
                                fontWeight: 700
                                color: '#1D1C21'
                          - Row:
                              gap: 5
                              children:
                                - Text:
                                    id: currentPercentage
                                    styles:
                                      fontWeight: 500
                                      fontSize: 14
                                      color: '#0E9F6E'
                                - Text:
                                    text: vs Last Month
                                    styles:
                                      fontWeight: 400
                                      fontSize: 14
                                      color: '#747184'
                    - Chart:
                        type: line
                        title: ''
                        labels:
```
Now the solution in this case is to provide the chart width of 40% as other child already has 60% space assigned. Becuase the default width for chart in my case is 365px so when collective size of container is lesser it just explodes and looks horrible. 
Any suggestions for this

packages/runtime/src/widgets/Search.tsx
```diff
@@ -57,34 +58,45 @@ export const Search: React.FC<SearchProps> = ({
...
+              : undefined,
+            backgroundColor: styles?.backgroundColor,
+            boxShadow: "none",
+          }}
+        />
+      </AutoComplete>
+      <style>
+        {`
+			/* Linear loader animation */
+			#${id ?? ""} {
```
evshi, line 94: Check for id block and just don't include style block if it's empty. Otherwise this is just extra elements.

packages/runtime/src/widgets/Search.tsx
```diff
@@ -57,34 +58,45 @@ export const Search: React.FC<SearchProps> = ({
...
-      />
-    </AutoComplete>
+    <div>
+      <AutoComplete
+        id={id}
+        allowClear
+        onSearch={handleSearch}
+        // TODO: Handle on search result select
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        onSelect={() => {}}
```
evshi, line 68: You can use `noop` from lodash

PR #179, opened 2023-11-16
added chartjs-datalabels-plugin support for stacked chart

sibteali786:
## Describe your changes
Added support for adding labels inside the stacked bar charts 
### Screenshots [Optional]
Screenshot 2023-11-16 at 7 42 37 PM

Closes #
This closes the [NAV2-49](https://jira.atlaslabs.work/browse/NAV2-49) JIRA ticket

- [ ] I have added tests

### DIFF
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -34,6 +34,7 @@
     "@types/react-dom": "^18.2.0",
     "antd": "^5.9.0",
     "chart.js": "^4.4.0",
+    "chartjs-plugin-datalabels": "^2.2.0",
     "eslint-config-custom": "workspace:*",
     "jest": "27.5.1",
     "lodash-es": "^4.17.21",
@@ -43,8 +44,5 @@
     "react-toastify": "^9.1.3",
     "tsconfig": "workspace:*",
     "tsup": "^7.2.0"
-  },
-  "dependencies": {
-    "chartjs-plugin-datalabels": "^2.2.0"
   }
 }
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -43,5 +43,8 @@
     "react-toastify": "^9.1.3",
     "tsconfig": "workspace:*",
     "tsup": "^7.2.0"
+  },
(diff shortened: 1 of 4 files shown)


packages/runtime/src/widgets/Charts/StackBarChart.tsx
```diff
@@ -80,6 +81,7 @@ export const StackBarChart: React.FC<BarChartProps> = (props) => {
           },
         },
       }}
+      plugins={[ChartDataLabels]}
```
evshi, line 84: Any reason not to add this to every Chart?

sibteali786: i ll have to test it thoroughly as even for Line Chart if no configuration is provided it just explodes the UI and it looks weird. I ll check their docs so that we can have properties users can use to switch it on and off.

packages/runtime/package.json
```diff
@@ -43,5 +43,8 @@
     "react-toastify": "^9.1.3",
     "tsconfig": "workspace:*",
     "tsup": "^7.2.0"
+  },
+  "dependencies": {
```
evshi, line 47: This should be in devDeps

packages/runtime/src/widgets/Charts/StackBarChart.tsx
evshi: Okay we'll just keep it for the stacked bar chart for now

packages/runtime/src/widgets/Charts/StackBarChart.tsx
```diff
@@ -1,6 +1,8 @@
 import { Bar } from "react-chartjs-2";
 import type { ChartOptions } from "chart.js";
 import type { Expression } from "@ensembleui/react-framework";
+// eslint-disable-next-line import/no-extraneous-dependencies
```
evshi, line 4: This comment shouldn't be needed since you're using it below

evshi (approved):

PR #184, opened 2023-11-22
Global block

sibteali786:
## Describe your changes
Allows dev to use Global Block to define their Functions and variables there.
#158 
Closes #
#158 

- [ ] I have added tests

### DIFF
--- a/apps/starter/src/ensemble/screens/dashboard.yaml
+++ b/apps/starter/src/ensemble/screens/dashboard.yaml
@@ -1,6 +1,6 @@
 View:
   onLoad:
-    invokeAPI:
+    invokeApi:
       name: getDummyJson
   header:
     title:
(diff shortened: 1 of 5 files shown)


sibteali786:
@anserwaseem if you also wanna review it, any suggestions from your side are also welcomed.

apps/starter/src/ensemble/screens/dashboard.yaml
```diff
@@ -1,6 +1,6 @@
 View:
   onLoad:
-    invokeApi:
+    invokeAPI:
```
evshi, line 3: I believe this is case sensitive so this would break things?

packages/framework/src/shared/models.ts
```diff
@@ -14,6 +14,7 @@ export interface EnsembleScreenModel {
   header?: EnsembleHeaderModel;
   footer?: EnsembleFooterModel;
   apis?: EnsembleAPIModel[];
+  global: string | undefined;
```
evshi, line 17: Use `?` for consistency

evshi (commented):
Small comments, but otherwise LGTM

evshi (approved):

PR #210, opened 2023-12-04
add ToggleButton widget

### LINKED ISSUE #175: Implement ToggleButton
ACs:

- [ ] Implement the ToggleButton schema

anserwaseem:

Closes #175

- [ ] I have added tests

### DIFF
--- a/apps/starter/src/ensemble/screens/askEmbrace.yaml
+++ b/apps/starter/src/ensemble/screens/askEmbrace.yaml
@@ -16,12 +16,45 @@ View:
   body:
     Column:
       children:
-        # - Markdown:
-        #     id: askText
-        #     text: "## Ask Embrace"
-        # - Text:
-        #     id: he
-        #     text: hehe
+        - Markdown:
+            id: askText
+            text: "### Toggle Buttons"
+
+        - ToggleButton:
+            id: myToggleButton
+            styles:
+              margin: 5
+              spacing: 10
+              runSpacing: 3
+              color: black
+              selectedColor: yellow
+              backgroundColor: white
+              selectedBackgroundColor: pink
+              selectedBorderColor: blue
+              borderColor: green
+            value: red
+            items:
+              # - red
+              # - green
+              # - blue
+              # - yellow
+              - { "label": "Red", "value": "red", "icon": { "name": "Star" } }
+              - {
+                  "label": "Green",
+                  "value": "green",
+                  "icon": { "name": "Add", "color": "blue", "size": 22 },
+                }
+              - { "label": "Yellow", "value": "yellow" }
+              - { "label": "Blue", "value": "blue" }
+              - { "label": "Orange", "value": "orange" }
+              # - { "label": "Purple", "value": "purple" }
+              # - { "label": "Pink", "value": "pink" }
+              # - { "label": "Brown", "value": "brown" }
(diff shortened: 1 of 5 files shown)


apps/starter/src/ensemble/screens/askEmbrace.yaml
```diff
@@ -16,12 +16,45 @@ View:
...
+              selectedBackgroundColor: pink
+              selectedBorderColor: blue
+              borderColor: green
+            value: red
+            items:
+              # - red
+              # - green
+              # - blue
+              # - yellow
+              - { "label": "Red", "value": "red", "icon": { "name": "Star" } }
```
evshi, line 41: This is equivalent to specifying a YAML object correct? This should still work if you have

```ts
- label: Red
  value: red
  icon:
    name: Star
```

packages/runtime/src/widgets/ToggleButton.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+export const ToggleButton: React.FC<ToggleButtonProps> = ({
+  id,
+  value: givenValue,
+  items,
+  onChange,
+  styles,
+}) => {
+  const [value, setValue] = useState(givenValue);
+  const { values } = useRegisterBindings({ value }, id, { setValue });
```
evshi, line 51: `styles` can now also be passed into evaluate expressions

packages/runtime/src/widgets/ToggleButton.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const action = useEnsembleAction(onChange);
+
+  const handleChange = (
```
evshi, line 54: Merge this into a single callback

packages/runtime/src/widgets/ToggleButton.tsx
```diff
@@ -0,0 +1,133 @@
...
+    setValue(newValue);
+    onChangeCallback(newValue);
+  };
+
+  const onChangeCallback = useCallback(
+    (newValue: string) => {
+      if (!action) {
+        return;
+      }
+      action.callback({
```
evshi, line 67: Only execute if the value is different?

packages/framework/src/parser.ts
```diff
@@ -197,17 +199,31 @@ export const unwrapWidget = (obj: Record<string, unknown>): EnsembleWidget => {
...
-          const unwrappedWidget = unwrapWidget(
-            widget as Record<string, unknown>,
-          );
-          return { label, icon, widget: unwrappedWidget };
-        },
-      );
-      set(properties as object, "items", valueItems);
+  if (!isEmpty(items) && isArray(items)) {
+    if (isString(items[0])) {
+      set(properties as object, "items", items);
```
evshi, line 204: We can always set the items property or leave it unchanged, and then just handle the special structure in your widget. Otherwise this will get very complicated trying to handle all the cases in the parser.

apps/starter/src/ensemble/screens/askEmbrace.yaml
anserwaseem: yes

evshi (approved):
Please run `pnpm changeset add` and commit a changeset so the packages will get versioned and published automatically

PR #237, opened 2023-12-11
Add Date widget

### LINKED ISSUE #58: Implement Date widget

anserwaseem:

Closes #58

- [ ] I have added tests


packages/runtime/src/widgets/Date.tsx
```diff
@@ -0,0 +1,361 @@
...
+
+  const { values } = useRegisterBindings({ ...props, value }, props.id, {
+    setValue,
+  });
+
+  const action = useEnsembleAction(props.onChange);
+  const onChangeCallback = useCallback(
+    (date: any) => {
+      const formattedDate = dayjs(toString(date))?.format(DateDisplayFormat);
+      formattedDate !== "Invalid Date" && setEnteredDate(formattedDate);
```
evshi, line 74: This should be a separate hook to sync the date with state. Hooks should be responsible for one thing

packages/runtime/src/widgets/Date.tsx
```diff
@@ -0,0 +1,361 @@
...
+          />
+        </LocalizationProvider>
+      </AntForm.Item>
+    </DatePickerContext.Provider>
+  );
+};
+
+WidgetRegistry.register("Date", Date);
+
+const CustomCalendarHeader: React.FC<PickersCalendarHeaderProps<any>> = (
```
evshi, line 207: Why is this extra component needed?

packages/runtime/src/widgets/Date.tsx
```diff
@@ -0,0 +1,361 @@
...
+              margin: "24px",
+            }}
+          />
+        )}
+      </div>
+    </>
+  );
+};
+
+const CustomActionBar: React.FC<PickersActionBarProps> = (props) => {
```
evshi, line 314: Why is this extra component needed?

packages/runtime/src/widgets/Date.tsx
anserwaseem: these extra components are being used here https://github.com/EnsembleUI/ensemble-react/blob/d6c200c43766f6c2db8c3064d8863df8e0fbd3f4/packages/runtime/src/widgets/Date.tsx#L130-L133

evshi (approved):

PR #243, opened 2023-12-13
exposed onItemSelect event

sibteali786:
## Describe your changes
Allows users to execute actions based on itemSelect change 

Closes #

- [ ] I have added tests


packages/runtime/src/widgets/Dropdown.tsx
```diff
@@ -18,16 +20,25 @@ interface SelectOption {
 
 const Dropdown: React.FC<DropdownProps> = (props) => {
   const [selectedValue, setSelectedValue] = useState<string>();
+  const handleChange = (value: string): void => {
```
evshi, line 23: not necessary to declare a new function here you can just pass the setSelectedValue directly to the prop

evshi (approved):

evshi:
Also @sibteali786 please run `pnpm changeset add` and provide a quick summary of the changes

PR #249, opened 2023-12-17
exposed index and lenght for iterating array

sibteali786:
## Describe your changes
- Exposed currentIndex and templateLength properties for item template
- Added property center for position in modal.
- A stylistic addition in textInput 
- Added Static Progress indication for Progress Bar and some properties addition 
- Support for Expressions in DataGrid header styles

Closes #

- [ ] I have added tests

### DIFF
--- a/apps/starter/src/ensemble/screens/createCollection.yaml
+++ b/apps/starter/src/ensemble/screens/createCollection.yaml
@@ -16,10 +16,9 @@ View:
         executeCode: |
           const name = collectionName.value
           const email = description.value
-          console.log(name, email)
-          //const res = ensemble.storage.get("arrayOfCollections")
-          //res.push({"id": res.length+1, "fields": { "title": name, "sources": email }})
-          //ensemble.storage.set("arrayOfCollections",res)
+          const res = ensemble.storage.get("arrayOfCollections")
+          res.push({"id": res.length+1, "fields": { "title": name, "sources": email }})
+          ensemble.storage.set("arrayOfCollections",res)
       children:
         - TextInput:
             id: collectionName
--- a/apps/starter/src/ensemble/screens/createCollection.yaml
+++ b/apps/starter/src/ensemble/screens/createCollection.yaml
@@ -16,9 +16,10 @@ View:
         executeCode: |
           const name = collectionName.value
           const email = description.value
-          const res = ensemble.storage.get("arrayOfCollections")
-          res.push({"id": res.length+1, "fields": { "title": name, "sources": email }})
-          ensemble.storage.set("arrayOfCollections",res)
+          console.log(name, email)
+          //const res = ensemble.storage.get("arrayOfCollections")
+          //res.push({"id": res.length+1, "fields": { "title": name, "sources": email }})
+          //ensemble.storage.set("arrayOfCollections",res)
       children:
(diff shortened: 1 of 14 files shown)


packages/runtime/src/widgets/Column.tsx
```diff
@@ -51,7 +51,16 @@ export const Column: React.FC<FlexboxProps> = (props) => {
...
       {childrenFirst ? renderedChildren : null}
       {namedData.map((n, index) => (
-        <CustomScopeProvider key={index} value={n as CustomScope}>
+        <CustomScopeProvider
+          key={index}
+          value={
+            {
+              ...n,
+              currentIndex: index,
+              templateLength: namedData.length,
```
evshi, line 60: Let's just keep this simple `index` and `length`

packages/runtime/src/widgets/Dropdown.tsx
```diff
@@ -47,6 +50,7 @@ const Dropdown: React.FC<DropdownProps> = (props) => {
       }}
     >
       <Select
+        defaultValue={String(values?.value)}
```
evshi, line 53: Not sure why this is needed?

sibteali786: defaultValue is needed in case of dropdown for enrollment screen to show a program which was being selected at selectDOS screen, it was not working before so i had to add it explicitly.

evshi: In that case, you would just set the value? Default value should only apply if no value is present.

packages/runtime/src/widgets/Tag.tsx
```diff
@@ -67,11 +67,11 @@ export const Tag: React.FC<TagProps> = (props) => {
...
+        fontWeight: values?.styles?.fontWeight
+          ? values?.styles.fontWeight
           : "normal",
-        fontFamily: props.styles?.fontFamily,
-        fontSize: props.styles?.fontSize ?? 12,
+        fontFamily: values?.styles?.fontFamily,
+        fontSize: values?.styles?.fontSize ?? 12,
         display: "inline-flex",
         alignItems: "center",
         margin: props.styles?.margin ?? "5px",
```
evshi, line 77: You can use spread operator to apply all the styles at once

evshi (approved):

apps/starter/src/ensemble/screens/createCollection.yaml
```diff
@@ -16,9 +16,10 @@ View:
         executeCode: |
           const name = collectionName.value
           const email = description.value
-          const res = ensemble.storage.get("arrayOfCollections")
-          res.push({"id": res.length+1, "fields": { "title": name, "sources": email }})
-          ensemble.storage.set("arrayOfCollections",res)
+          console.log(name, email)
```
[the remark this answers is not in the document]
sibteali786: nope only for checking something, i can revert if that important

packages/runtime/src/widgets/Tag.tsx
sibteali786: agree, let me refactor it

sibteali786: @evshi  updated it

PR #259, opened 2023-12-20
252 implement social login widget2

sagardspeed2:
Functionality for the google login
Created a custom widget for google login

- [ ] I have added tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "prettier": "^3.0.2",
     "ts-jest": "27.1.5",
     "tsconfig": "workspace:*",
-    "turbo": "1.9.9"
+    "turbo": "1.11.2"
   },
   "name": "ensemble-react",
   "packageManager": "pnpm@6.32.2",
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "prettier": "^3.0.2",
     "ts-jest": "27.1.5",
     "tsconfig": "workspace:*",
-    "turbo": "1.11.2"
+    "turbo": "1.9.9"
   },
   "name": "ensemble-react",
   "packageManager": "pnpm@6.32.2",
--- a/packages/framework/src/hooks/useApplicationContext.tsx
+++ b/packages/framework/src/hooks/useApplicationContext.tsx
@@ -23,7 +23,7 @@ export const ApplicationContextProvider: React.FC<
         appContext={{
           ...defaultApplicationContext,
           application: app,
-          env: app.config?.environmentVariables ?? null,
+          env: app.config?.environmentVariables ?? {},
         }}
       >
         {children}
--- a/packages/framework/src/hooks/useApplicationContext.tsx
+++ b/packages/framework/src/hooks/useApplicationContext.tsx
@@ -23,6 +23,7 @@ export const ApplicationContextProvider: React.FC<
         appContext={{
           ...defaultApplicationContext,
           application: app,
+          env: app.config?.environmentVariables ?? null,
         }}
       >
         {children}
--- a/packages/framework/src/shared/dto.ts
+++ b/packages/framework/src/shared/dto.ts
@@ -3,6 +3,7 @@
  *
  * Mostly equivalent to raw JSON definition over wire
  */
+
 export interface ApplicationDTO {
   name: string;
   id: string;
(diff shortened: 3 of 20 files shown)


packages/runtime/src/EnsembleApp.tsx
```diff
@@ -28,14 +30,18 @@ export const EnsembleApp: React.FC<EnsembleAppProps> = ({
   appId,
   application,
 }) => {
+  // TODO: need to move this inside env
+  const oAuthClientId =
+    "726646987043-9i1it0ll0neojkf7f9abkagbe66kqe4a.apps.googleusercontent.com";
```
evshi, line 35: You can stub the env property in the application context for now and move this to the starter app

packages/runtime/src/EnsembleApp.tsx
```diff
@@ -62,13 +68,23 @@ export const EnsembleApp: React.FC<EnsembleAppProps> = ({
...
+    [app]
   );
+
   return (
     <ApplicationContextProvider app={app}>
       <ThemeProvider>
-        <RouterProvider router={router} />
-        <ToastContainer />
+        {oAuthClientId ? (
+          <GoogleOAuthProvider clientId={oAuthClientId}>
```
evshi, line 78: Have you tried just moving this into the component? Looking through the source for this library, I don't see any reason why this has to be at the top of the app

package.json
```diff
@@ -17,7 +17,7 @@
     "prettier": "^3.0.2",
     "ts-jest": "27.1.5",
     "tsconfig": "workspace:*",
-    "turbo": "1.9.9"
+    "turbo": "1.11.2"
```
evshi, line 20: Please don't update this

packages/runtime/package.json
```diff
@@ -46,5 +46,9 @@
     "react-use": "^17.4.2",
     "tsconfig": "workspace:*",
     "tsup": "^7.2.0"
+  },
+  "dependencies": {
+    "@react-oauth/google": "^0.12.1",
```
evshi, line 51: These should be moved to dev dependencies

packages/runtime/src/widgets/SignInWithGoogle.tsx
```diff
@@ -0,0 +1,88 @@
...
+  text?: "signin_with" | "signup_with";
+  shape?: "rectangular" | "pill" | "circle" | "square";
+  onSignedIn?: EnsembleAction;
+  onError?: EnsembleAction;
+};
+
+export const SignInWithGoogle: React.FC<SignInWithGoogleProps> = (props) => {
+  // TODO: need to move this inside env
+  const oAuthClientId =
+    "726646987043-9i1it0ll0neojkf7f9abkagbe66kqe4a.apps.googleusercontent.com";
```
evshi, line 24: If you don't need this at the application context provider level, this can just be a prop in your widget.

evshi (changes requested):

packages/framework/src/hooks/useApplicationContext.tsx
```diff
@@ -23,6 +23,7 @@ export const ApplicationContextProvider: React.FC<
         appContext={{
           ...defaultApplicationContext,
           application: app,
+          env: app.config?.environmentVariables ?? null,
```
evshi, line 26: Not sure why the nullish coalescing is needed `??` ?

evshi, line 26: Not sure why the nullish coalescing is needed `??` ?

packages/framework/src/shared/dto.ts
```diff
@@ -34,3 +36,11 @@ export interface ScriptDTO {
   name: string;
   content: string;
 }
+
+export interface EnsembleEnvironmentDTO {
```
evshi, line 40: This should just be a generic Record<string unknown> type since the user can put any environment variables they want.

packages/runtime/package.json
```diff
@@ -46,5 +47,8 @@
     "react-use": "^17.4.2",
     "tsconfig": "workspace:*",
     "tsup": "^7.2.0"
+  },
+  "dependencies": {
+    "jwt-decode": "^4.0.0"
```
evshi, line 52: This should also be in devDeps

evshi (approved):

PR #272, opened 2023-12-27
Fix chart options

anserwaseem:

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Charts/BarChart.tsx
+++ b/packages/runtime/src/widgets/Charts/BarChart.tsx
@@ -2,7 +2,8 @@ import { Bar } from "react-chartjs-2";
 import type { ChartOptions } from "chart.js";
 import { useRegisterBindings } from "@ensembleui/react-framework";
 import { useState } from "react";
-import { getMergedOptions, type ChartDataSets, type ChartProps } from "..";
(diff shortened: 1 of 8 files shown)


packages/runtime/src/widgets/Charts/BarChart.tsx
```diff
@@ -39,6 +40,19 @@ export const BarChart: React.FC<ChartProps> = (props) => {
     setTitle,
   });
 
+  merge(
```
evshi, line 43: Can we do this in the Chart index component to avoid repeating for each chart?

packages/runtime/src/widgets/Charts/LineChart.tsx
```diff
@@ -46,19 +47,7 @@ export const LineChart: React.FC<ChartProps> = (props) => {
...
-        ...(config?.options as ChartOptions<"line">),
-        plugins: {
-          title: {
-            display: Boolean(values?.title),
-            text: values?.title,
-          },
-          legend: {
-            display: false,
-          },
-        },
```
evshi, line 60: This needs to go in default options or will create regression?

evshi (approved):

PR #274, opened 2023-12-28
history and location api

### LINKED ISSUE #174: Location and history API

**Additional context**

sagardspeed2:
Feat: History api - ensemble.history
WIP: location api

## Describe your changes
Added location and history API

closes #174 

- [ ] I have added tests


packages/runtime/src/runtime/entry.tsx
```diff
@@ -3,13 +3,19 @@ import { Outlet, useNavigate, useLocation } from "react-router-dom";
...
 
 interface EnsembleEntryProps {
   entry: EnsembleEntryPoint;
 }
 export const EnsembleEntry: React.FC<EnsembleEntryProps> = ({ entry }) => {
   const location = useLocation();
   const navigate = useNavigate();
+
+  // store copy of navigate and location for the use of outside components
+  browserHistory.navigate = navigate;
```
evshi, line 16: The react router navigate function alone is not sufficient. We also need ability to pass inputs and validate the screen etc. You need to replicate/refactor the useNavigateScreen hook

packages/framework/src/state/user.ts
```diff
@@ -20,3 +20,7 @@ export const useEnsembleUser = (): [
   const [user, setUser] = useAtom(userAtom);
   return [user, setUser];
 };
+
+export const useEnsembleHistory = (): History => {
```
evshi, line 24: What is this for?

packages/runtime/src/runtime/history.tsx
```diff
@@ -0,0 +1,11 @@
+import { type NavigateFunction, type Location } from "react-router-dom";
+
+export interface BrowserHistoryDefinition {
+  navigate: NavigateFunction | unknown;
+  location: Location | unknown;
+}
+
+export const browserHistory: BrowserHistoryDefinition = {
```
evshi, line 8: You can use an atom for this don't just store references globally they can create issues.

evshi (changes requested):

packages/framework/src/state/user.ts
sagardspeed2: That is only for testing, and forgot to remove it before commit

packages/runtime/src/runtime/entry.tsx
sagardspeed2: Done

packages/runtime/src/runtime/navigateApi.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { findExpressions, evaluate } from "@ensembleui/react-framework";
+import { isString, cloneDeep, set } from "lodash-es";
+import type { NavigateFunction, Location } from "react-router-dom";
+
+export interface BrowserHistoryDefinition {
+  navigate: NavigateFunction | undefined;
+  location: Location | undefined;
+}
+
+export const browserHistory: BrowserHistoryDefinition = {
```
evshi, line 16: Don't use a global object

packages/runtime/src/runtime/navigateApi.tsx
```diff
@@ -0,0 +1,65 @@
...
+    const expressionMap: string[][] = [];
+    findExpressions(inputs, [], expressionMap);
+    expressionMap.forEach(([path, value]) => {
+      const result = evaluate(screenContext, value, {
+        ensemble: { storage },
+        ...customScope,
+      });
+      set(inputs, path, result);
+    });
+  }
```
evshi, line 55: Does this not throw errors for you in console? This is being evaluated in the context of the executeCode scope so you can only reference the context being passed in. You don't need to evaluate these because they will be automatically dereferenced in the code block

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -95,7 +100,15 @@ export const useExecuteCode: EnsembleActionHook<
...
           js,
           merge(
-            { ensemble: { storage, formatter, env: appContext?.env } },
+            {
+              ensemble: {
+                storage,
+                formatter,
+                env: appContext?.env,
+                navigateScreen: (targetScreen: NavigateScreenAction) =>
+                  navigateApi(targetScreen, screen, customScope, storage),
```
evshi, line 109: Where is the location API?

packages/runtime/src/runtime/navigateApi.tsx
sagardspeed2: No, this is not throwing any errors in the console. also whatever screen name user passed with the ensemble.navigateScreen API, first we need to validate that screen name is the right one and that screen is exists within the app scope. that's why I need to evaluate this. I am passing the hook context from the useExecuteCode block  to navigateApi, in order to access those hooks.

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
sagardspeed2: I supposed this navigateScreen is alias of location api, for move from one page to other page inside the useExecuteCode API

evshi: The location and history APIs should be separate members. We need a way to view the current location and history.

packages/runtime/src/runtime/navigateApi.tsx
evshi: Hmm then the matching screen logic can stay, but the input/expression evaluation doesn't need to be here.

packages/runtime/src/runtime/navigateApi.tsx
sagardspeed2: I had idea about the using atom, but then I realized that the jotai is not used in the runtime repo. So I used like this.

sagardspeed2: And I think this will not cause any issue and useful in other APIs as well, where we need the access of the useNavigate.

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
sagardspeed2: okay should I consider this navigate API as a history API ?

sagardspeed2: @evshi As per the react router v6, history is removed from the react router dom. and also we can't get the history stack from the window.history as per the security related concern. because the history is communicate with the tab, not the specific domain.

packages/runtime/src/runtime/history.tsx
```diff
@@ -0,0 +1,11 @@
+import type { NavigateFunction, Location } from "react-router-dom";
+
+export interface BrowserHistoryDefinition {
+  navigate: NavigateFunction | undefined;
+  location: Location | undefined;
+}
+
+export const browserHistory: BrowserHistoryDefinition = {
```
evshi, line 8: @sagardspeed2 please get rid of this global object

evshi (commented):
Remove the global object and instantiate the navigate and location APIs at point of use.

evshi (approved):

PR #277, opened 2023-12-29
make menu bar responsive on small screen sizes

### LINKED ISSUE #30: Make menu bar responsive on small screen sizes
ACs:
- MenuBar should automatically move from side to top for small screen sizes
- MenuBar should support collapsible callback
- Collapsed state should only show icons of menu items

anserwaseem:

Closes #30

- [ ] I have added tests

### DIFF
--- a/apps/starter/src/ensemble/screens/home.yaml
+++ b/apps/starter/src/ensemble/screens/home.yaml
@@ -5,14 +5,17 @@ ViewGroup:
       width: 240px
       iconWidth: 20px
       iconHeight: 20px
-      backgroundColor: "#fcfeff"
-      labelColor: "#171B2A"
-      selectedColor: "#171B2A"
+      backgroundColor: lime
+      labelColor: blue
       labelFontSize: 1
-      headerCollapsedWidth: 160px
+      borderRadius: 10px
       onSelectStyles:
-        backgroundColor: "#D5DAEA"
-        borderRadius: 10px
+        iconWidth: 30px
+        iconHeight: 30px
+        backgroundColor: red
+        borderRadius: 20px
+        labelColor: green
+        labelFontSize: 2
     header:
       # Markdown:
       #   text: "# Ensemble"
@@ -43,7 +46,8 @@ ViewGroup:
     footer:
       Text:
         text: Give us feedback
+    isCollapsible: true
     onCollapse:
       executeCode:
         body: |-
-          console.log('mySidebar', mySidebar);
+          console.log('sidebar collapsed', mySidebar);
--- a/apps/starter/src/ensemble/screens/home.yaml
+++ b/apps/starter/src/ensemble/screens/home.yaml
@@ -1,5 +1,6 @@
 ViewGroup:
   SideBar:
+    id: mySidebar
     styles:
       width: 240px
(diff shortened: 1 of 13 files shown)


apps/starter/src/ensemble/screens/home.yaml
```diff
@@ -38,3 +43,7 @@ ViewGroup:
     footer:
       Text:
         text: Give us feedback
+    onCollapse:
```
evshi, line 46: Sorry, this should have been clarified in the ticket more. Instead of `onCollapse`, we should support something like
```
Button:
  onTap:
    mySidebarMenu.collapse()
```

Basically  expose a way to collapse the menu programmatically so users can put their own UI over it also

packages/runtime/src/runtime/menu.tsx
```diff
@@ -216,9 +296,47 @@ export const SideBarMenu: React.FC<MenuBaseProps> = (props) => {
...
+        }}
+      >
+        {renderMuiIcon(
+          props.collapsed
+            ? "KeyboardDoubleArrowRight"
+            : "KeyboardDoubleArrowLeft",
+          "24px",
+          "24px",
+        )}
+      </Icon>
```
evshi, line 329: Let's expose a separate property `isCollapsible` for this

packages/runtime/src/runtime/menu.tsx
```diff
@@ -36,27 +46,32 @@ interface MenuItem {
...
+}
+
+type MenuBaseProps = {
+  items: MenuItem[];
   header?: EnsembleWidget;
   footer?: EnsembleWidget;
   enableSearch?: boolean;
-}
+  onCollapse?: EnsembleAction;
+  collapsed: boolean;
```
evshi, line 72: nit: use `isCollapsed`

packages/runtime/src/runtime/menu.tsx
```diff
@@ -36,27 +46,32 @@ interface MenuItem {
...
-    labelFontSize?: number;
-    searchBoxColor?: TypeColors;
-    iconWidth?: string;
-    iconHeight?: string;
-    width?: string;
-    onSelectStyles?: {
-      backgroundColor?: TypeColors;
-      borderRadius?: string;
-    };
+    borderRadius?: string;
```
evshi, line 62: Let's clean this up a bit better - consolidate the selected styles in `onSelectStyles` and remove irrelevant styles like `searchBoxColor`.

packages/runtime/src/runtime/menu.tsx
```diff
@@ -36,27 +46,32 @@ interface MenuItem {
...
+  backgroundColor?: TypeColors;
+  labelColor?: TypeColors;
+  selectedColor?: TypeColors;
+  labelFontSize?: number;
+  searchBoxColor?: TypeColors;
+  iconWidth?: string;
+  iconHeight?: string;
+  width?: string;
+  headerCollapsedWidth?: string;
+  footerCollapsedWidth?: string;
```
evshi, line 59: These are probably not needed? In collapsed state the header/footer can use expressions to render a different element

evshi:
Closed for #363

PR #289, opened 2024-01-05
fix: support expression evaluate inside placeholder of search widget

### LINKED ISSUE #288: Search bar properties are not all expressions
**Describe the bug**
Search bar does not use expressions for all of its properties

**To Reproduce**

1. Go to screen with Search
2. Try using expression in placeholder
3. Text is just the expression

**Expected behavior**
Expression should be evaluated

**Screenshots**

**Additional context**

sagardspeed2:
## Describe your changes
Support expression evaluation inside placeholder string of search widgets

Closes #
closes #288

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Search.tsx
+++ b/packages/runtime/src/widgets/Search.tsx
@@ -54,7 +54,7 @@ export const Search: React.FC<SearchProps> = ({
     name: itemTemplate?.name,
   });
   const { rootRef, values } = useRegisterBindings(
-    { styles, value, options },
+    { styles, value, options, placeholder },
     id,
     {
       setValue,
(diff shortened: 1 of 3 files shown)


packages/runtime/src/widgets/Search.tsx
```diff
@@ -54,7 +54,7 @@ export const Search: React.FC<SearchProps> = ({
     name: itemTemplate?.name,
   });
   const { rootRef, values } = useRegisterBindings(
-    { styles, value, options },
+    { styles, value, options, placeholder },
```
evshi, line 57: Let's destructure the props for the props that shouldn't be expressions but then otherwise pass everything in as bindings, so be inclusive instead of exclusive. There's a few examples of this like `Column`. This avoids us having to update it further for every new property.

evshi (approved):

PR #290, opened 2024-01-05
287 support endpoint apis in custom widget

### LINKED ISSUE #287: Support APIs in custom widget definitions
In web dev, it is common to associate a component with an endpoint so the functionality is portable, e.g. Search bar.

Allow API definitions in custom widgets and allow them to reference/invoke them.

**Additional context**

sagardspeed2:
## Describe your changes
Added enspoint API support in custom widgets

Closes #
closes #287 

- [ ] I have added tests


evshi (changes requested):
Instead of creating a separate state for widget APIs, let's just combine all the APIs at the parsing stage.

sagardspeed2:
@evshi, I am suggesting that we can create a new separate state for the APIs and combine all the APIs from all screens and custom widgets into that state. We can then remove the APIs from the screen state.

evshi:
@sagardspeed2 API definitions are scoped to screens. A screen cannot reference the API of another screen unless it's explicitly exported/stored. The API state is also scoped to the screen lifecycle. It doesn't make sense for the loading and response states to persist beyond the screen.

sagardspeed2:
@evshi, so, based on what you're suggesting, how should we go about combining all the APIs from screen and custom widgets?

evshi:
Concatenate the API definition of widgets to the screen API definitions during parsing

sagardspeed2:
@evshi we can not combine the custom widget APIs into screen apis. because of,

while parsing,
we are registering the custom widgets into widget registry before the screen will be passed to screencontextprovider.
and the screen atom will update when screen will change.

so when user will change the screen, then the screen atom will be updated with the new screen and thus APIs inside the screen atom will also be replaced by the new screen apis.

evshi:
@sagardspeed2 that's not completely correct. The lifecycle for the API models goes like this:

1. Parsing - all this does is convert YAML into JS objects. This is where you can take the widget API objects and concatenate them with the screen API object array.
2. Initialization - the screen atom is set with the screen model, which includes all the api models. the concatenation should happen before this
3. Invocation (can be done multiple times) - during runtime, the api model is looked up and the invocation state is saved in the `data` block of the screen. this state can be mutated multiple times in a screen
5. Teardown - when the screen is unloaded, the atom is reset and all API models are removed until the new screen is loaded

packages/framework/src/parser.ts
```diff
@@ -46,22 +47,28 @@ export interface EnsembleWidgetYAML {
...
+
     const screens = app.screens.map(({ name, content: yaml }) => {
       const screen = parse(yaml) as EnsembleScreenYAML;
       const viewGroup = get(screen, "ViewGroup");
       if (viewGroup) {
         return EnsembleParser.parseMenu(viewGroup);
       }
-      return EnsembleParser.parseScreen(name, screen, app);
+
+      return EnsembleParser.parseScreen(name, screen, app, widgetApis);
```
evshi, line 66: Concat after instead of introducing another argument

apps/starter/src/ensemble/widgets/Header.yaml
```diff
@@ -29,7 +29,7 @@ Widget:
             searchKey: email
             onSearch:
               invokeApi:
-                name: searchUsers
+                name: findUsers
```
evshi, line 32: Can you add example in kitchen sink custom widgets page?

packages/framework/src/parser.ts
sagardspeed2: @evshi are you suggesting like, 
`return EnsembleParser.parseScreen(name, screen, {app, widgetApis});`

evshi (approved):

PR #299, opened 2024-01-09
292 support import script inside global scope

### LINKED ISSUE #292: Support import script in Global block
We don't have javascript highlighting in YAML files. one way is to break out the javascript into a separate file and import it via yaml. This is already supported in executeCode function.

Update the global block to also support scriptName and import file from app context

**Additional context**

sagardspeed2:
## Describe your changes
handle script import while the parser of script

Closes #
closes #292 

- [ ] I have added tests


apps/starter/.eslintrc.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   extends: ["custom/react"],
   rules: {
     "unicorn/filename-case": "off",
+    "no-console": "off",
```
evshi, line 5: Please remove

apps/starter/src/ensemble/screens/dashboard.yaml
```diff
@@ -0,0 +1,425 @@
+View:
```
evshi, line 1: Do not add more screens to starter project. Add your example usage to the kitchen sink

apps/starter/src/ensemble/scripts/hello.js
```diff
@@ -1,4 +1,9 @@
 console.log("hello data manipulation! check out the response below:");
 console.log(response);
 
+// eslint-disable-next-line no-unused-vars
+function callMeAsYouSee() {
+  console.log("See ya it works");
+}
+
```
evshi, line 8: Do not add this to starter app

packages/framework/src/parser.ts
```diff
@@ -114,7 +115,14 @@ export const EnsembleParser = {
...
     const apis = unwrapApiModels(screen);
 
-    const global = get(screen, "Global");
+    const globalBlock = get(screen, "Global");
+    const scriptName = get(globalBlock, "scriptName");
+    let global: string | undefined = globalBlock;
+    if (scriptName) {
+      global = app.scripts?.find((script) => script.name === scriptName)
+        ?.content;
+    }
```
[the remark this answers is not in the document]
sagardspeed2: Hey, so here is all needed validation is added. If global block is not contain the scriptName, then it will be use a string from the global scope and even if there will be no script file existence while passing the scriptName, then it will become undefined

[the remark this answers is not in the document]
evshi: What if global block is not a string? What if scriptName is not a string?

packages/framework/src/parser.ts
```diff
@@ -114,7 +116,18 @@ export const EnsembleParser = {
...
+    let global: string | undefined;
+    if (!isString(globalBlock)) {
+      const scriptName = get(globalBlock, "scriptName");
+      if (isString(scriptName)) {
+        global = app.scripts?.find((script) => script.name === scriptName)
+          ?.content;
+      }
+    } else if (isString(globalBlock)) {
+      global = globalBlock;
+    }
```
evshi, line 129: nit: can be simplified

```
if (isString(scriptName)) {
 ...
}
if (isEmpty(global) && isString(globalBlock)) {
...
}
```

evshi (approved):

PR #301, opened 2024-01-10
292 Allow passthrough  of HTML attributes

### LINKED ISSUE #293: Allow pass through of HTML attributes to node
HTML attributes are important for browser compatibilty and accessibility. We should support allowing customers to set custom HTML attrs.

Add a new property for `htmlAttributes` that will pass through to the HTML node.

**Additional context**

sagardspeed2:
## Describe your changes
Support passthrough of HTML attributes

Closes #
closes #293 

- [ ] I have added tests

### DIFF
--- a/apps/kitchen-sink/src/ensemble/screens/home.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/home.yaml
@@ -11,6 +11,8 @@ View:
         - Text:
             styles:
               names: heading-1
+            htmlAttributes:
+              data-value: hello
             text: The Kitchen Sink
         - Markdown:
             text: This application is built with Ensemble and serves as a guide for how to implement common UI patterns with Ensemble React.
--- a/packages/runtime/src/shared/types.ts
+++ b/packages/runtime/src/shared/types.ts
@@ -13,6 +13,7 @@ export interface EnsembleWidgetProps<
 > {
   id?: string;
   styles?: T;
+  htmlAttributes?: Record<string, unknown>;
   [key: string]: unknown;
 }
 
--- a/packages/runtime/src/widgets/Text.tsx
+++ b/packages/runtime/src/widgets/Text.tsx
@@ -35,6 +35,7 @@ export const Text: React.FC<TextProps> = (props) => {
       className={values?.styles?.names}
       ref={rootRef}
       style={{ ...values?.styles }}
+      {...values?.htmlAttributes}
     >
       {values?.text}
     </Typography.Text>


sagardspeed2:
@evshi check this PR, this is for the demo purpose only. Is this the right for passing HTML attributes ?

apps/kitchen-sink/src/ensemble/screens/home.yaml
```diff
@@ -11,6 +11,9 @@ View:
         - Text:
             styles:
               names: heading-1
+            htmlAttributes:
+              Value: hello
+              type: kio
```
evshi, line 16: Can you add a new section in Custom Widgets page for this? Title: Adding HTML Attributes

packages/framework/src/hooks/useHtmlPassThrough.tsx
```diff
@@ -0,0 +1,34 @@
+import type { RefCallback } from "react";
+import { isEmpty } from "lodash-es";
+import { useCallback } from "react";
+import type { Expression } from "../shared";
+
+export const useHtmlPassThrough = (
+  resolvedTestId: string | undefined,
+  resolvedWidgetId: string,
+  id?: Expression<string>,
```
evshi, line 9: Combine this into a single value for testId.

packages/framework/src/hooks/useHtmlPassThrough.tsx
```diff
@@ -0,0 +1,34 @@
...
+  htmlAttributes?: Record<string, string>,
+): { rootRef: RefCallback<never> } => {
+  const rootRef = useCallback(
+    (node: never) => {
+      if (node && "setAttribute" in node) {
```
evshi, line 14: Return early if no node or setAttribute fn

packages/framework/src/hooks/useHtmlPassThrough.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+        if (!isEmpty(htmlAttributes)) {
+          Object.keys(htmlAttributes).forEach((key: string) => {
+            (node as HTMLElement).setAttribute(
+              `data-${key}`,
+              htmlAttributes[key],
+            );
+          });
+        }
+      }
```
evshi, line 28: Apply this first so data-testid can override.

packages/framework/src/hooks/useHtmlPassThrough.tsx
```diff
@@ -0,0 +1,34 @@
...
+      if (node && "setAttribute" in node) {
+        (node as HTMLElement).setAttribute(
+          "data-testid",
+          id ? resolvedWidgetId : resolvedTestId ?? "",
+        );
+
+        if (!isEmpty(htmlAttributes)) {
+          Object.keys(htmlAttributes).forEach((key: string) => {
+            (node as HTMLElement).setAttribute(
+              `data-${key}`,
```
evshi, line 23: Don't include data prefix

packages/framework/src/hooks/useRegisterBindings.ts
```diff
@@ -23,10 +32,23 @@ export const useRegisterBindings = <T extends Record<string, unknown>>(
   methods?: InvokableMethods,
 ): RegisterBindingsResult<T> => {
   const testId = get(values, ["testId"]);
-  const { resolvedWidgetId, rootRef } = useWidgetId(
+  let htmlAttributes = get(values, "htmlAttributes") as Record<string, string>;
+
+  if (!isEmpty(htmlAttributes) && !isString(htmlAttributes)) {
+    htmlAttributes = mapKeys(htmlAttributes, (_, key) => key.toLowerCase());
+  }
```
evshi, line 39: This logic should be in the hook itself

packages/framework/src/hooks/useRegisterBindings.ts
```diff
@@ -23,10 +32,23 @@ export const useRegisterBindings = <T extends Record<string, unknown>>(
...
+
+  const { resolvedWidgetId, resolvedTestId } = useWidgetId(
     id,
     isString(testId) ? String(testId) : undefined,
   );
+  const { rootRef } = useHtmlPassThrough(
+    resolvedTestId,
+    resolvedWidgetId,
+    id,
+    htmlAttributes,
```
evshi, line 49: Take the html attributes from the values so they can be evaluated as bindings.

packages/runtime/src/shared/types.ts
```diff
@@ -13,6 +13,7 @@ export interface EnsembleWidgetProps<
 > {
   id?: string;
   styles?: T;
+  htmlAttributes?: Record<string, unknown>;
```
evshi, line 16: Value type is Expression<string>

evshi (commented):
Can you also add a changeset please

packages/framework/src/hooks/useRegisterBindings.ts
sagardspeed2:
```
const htmlAttributes = get(values, "htmlAttributes") as Record<
    string,
    string
  >;
  ```
  
  I take a htmlAttributes from the values using get method. are you suggesting `values.htmlAttributed` ?

Comment (evshi) [reply packages/framework/src/hooks/useRegisterBindings.ts:49]: Sorry, I meant the new values, after the bindings are resolved and combined.

File: packages/runtime/src/shared/types.ts (hunk 8 above)
Comment (sagardspeed2) [reply packages/runtime/src/shared/types.ts:16]: @evshi here htmlAttributes will be the object with the key value pair. so is this right ?
`htmlAttributes?: Record<string, Expression<string>>`

File: packages/framework/src/hooks/useHtmlPassThrough.tsx [hunk 9]
```diff
@@ -0,0 +1,29 @@
...
+import { useCallback } from "react";
+
+export const useHtmlPassThrough = (
+  widgetId: string,
+  htmlAttributes?: Record<string, string>,
+): { rootRef: RefCallback<never> } => {
+  const rootRef = useCallback(
+    (node: never) => {
+      if (node && "setAttribute" in node) {
+        if (!isEmpty(htmlAttributes) && !isString(htmlAttributes)) {
```
evshi, line 12: FYI there is an isObject util

packages/framework/src/hooks/useHtmlPassThrough.tsx
```diff
@@ -0,0 +1,29 @@
+import type { RefCallback } from "react";
+import { isEmpty, isString, mapKeys } from "lodash-es";
+import { useCallback } from "react";
+
+export const useHtmlPassThrough = (
+  widgetId: string,
```
evshi, line 6: Let's rename this to testId, make it nullable, and second param.

evshi (approved):
Some small comments but overall looks good

packages/framework/src/hooks/useHtmlPassThrough.tsx
sagardspeed2: @evshi what you mean by  second param ?

Issue #264, opened 2023-12-22
Expressions in one Screen break item-template data on another screen


sibteali786:
**Describe the bug**
- when we have expressions in SideBar Widget, like for any color or fontSize etc. 
- This breaks the patients screen such that 
1. Every patient card shows same task info. as in SS ( before and after )
2. Then when i move to enrollment screen the patient details do not work anymore becuase a fucntion ( used for handling a task user select at SelectDOS ) break becuase task is no longer an array.

**To Reproduce**
Simply go to SideBar screen in atlas and change any widget style property like color, fontSize to an expression, then we can simply observe the issue as mentioned in description.

**Expected behavior**

**Screenshots**
https://media.discordapp.net/attachments/1106512057555226634/1187765969460924550/Screenshot_2023-12-22_at_7.34.53_PM.png?ex=65981442&is=65859f42&hm=4da9ff317c83dee4aa600e46c7e8852c52d0ce922f63b0da08ce6b5089285798&=&format=webp&quality=lossless&width=1348&height=982

https://media.discordapp.net/attachments/1106512057555226634/1187765969460924550/Screenshot_2023-12-22_at_7.34.53_PM.png?ex=65981442&is=65859f42&hm=4da9ff317c83dee4aa600e46c7e8852c52d0ce922f63b0da08ce6b5089285798&=&format=webp&quality=lossless&width=1348&height=982

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

sagardspeed2:
Issue related nested item-template is not capturing the data from the parent customscopeprovider

sibteali786:
Thanks for mentioning it here @sagardspeed2 i ll take it from here

evshi:
Root cause is useTemplateData does not import custom scope. Similar issue with Chart config. @sibteali786 do you have scope to work on it? If not let's pass back to @sagardspeed2

PR #302, opened 2024-01-10
Item template fix

sibteali786:
## Describe your changes
- Fixed nested item template scope having no access to the parent scope.
- Fixed createBindingAtom to cause few properties like screen inputs to go blank when we
#264 
Closes #
#264

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/hooks/useTemplateData.ts
+++ b/packages/framework/src/hooks/useTemplateData.ts
@@ -38,6 +38,7 @@ export const useTemplateData = ({
     // Support legacy case where we don't always require curly braces
     return createBindingAtom(`\${${String(data)}}`);
   }, [data, isDataString, scope]);
+  console.log("From Template", scope);
   const rawData = useAtomValue(dataAtom);
   const namedData = useMemo(
     () => map(rawData, (value: unknown) => ({ [name]: value })),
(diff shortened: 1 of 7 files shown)


packages/framework/src/hooks/useTemplateData.ts
```diff
@@ -26,22 +27,23 @@ export const useTemplateData = ({
...
     }
     if (isExpression(data)) {
-      return createBindingAtom(data);
+      return createBindingAtom(data, scope);
     }
     // Support legacy case where we don't always require curly braces
     return createBindingAtom(`\${${String(data)}}`);
-  }, [data, isDataString]);
+  }, [data, isDataString, scope]);
+  console.log("From Template", scope);
```
evshi, line 41: Remove

packages/framework/src/state/widget.ts
```diff
@@ -102,12 +100,21 @@ export const createBindingAtom = <T = unknown>(
...
     });
+    let mergedInputs = {};
+    if (context) {
+      mergedInputs = {
+        ...(inputs as object),
+        ...context,
+      };
+    } else {
+      mergedInputs = inputs as object;
+    }
```
evshi, line 111: I'm not following why this is needed? Context should override input properties in the merge below?

evshi (commented):
Instead of repeating the custom scope hook twice in the template hook and the component, we can have the template hook return the new combined scope.

PR #306, opened 2024-01-11
added onClick event to Row widget

sibteali786:
## Describe your changes
- Added onTap event to Row widget 
- Modified hover behavior to pointer

Closes #

- [ ] I have added tests

### DIFF
--- a/apps/kitchen-sink/src/ensemble/screens/home.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/home.yaml
@@ -14,20 +14,6 @@ View:
             text: The Kitchen Sink
         - Markdown:
             text: This application is built with Ensemble and serves as a guide for how to implement common UI patterns with Ensemble React.
-        - Row:
-            onTap:
-              executeCode:  |
-                console.log('Row Clicked');
-            children:
-              - Text:
-                  text: Testing the Row 
-        - Column:
-            onTap:
-              executeCode:  |
-                console.log('Column Clicked');
-            children:
-              - Text:
-                  text: Testing the Column 
 
 Global:
   scriptName: hello.js
--- a/apps/kitchen-sink/src/ensemble/screens/home.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/home.yaml
@@ -21,6 +21,13 @@ View:
             children:
               - Text:
                   text: Testing the Row 
+        - Column:
+            onTap:
+              executeCode:  |
+                console.log('Column Clicked');
+            children:
+              - Text:
+                  text: Testing the Column 
 
 Global:
   scriptName: hello.js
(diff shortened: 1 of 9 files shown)


packages/runtime/src/widgets/Row.tsx
```diff
@@ -51,8 +61,10 @@ export const Row: React.FC<FlexboxProps> = (props) => {
         flexDirection: "row",
         flexFlow: "unset",
         flexGrow: "unset",
+        cursor: isHovered ? "pointer" : "auto",
```
evshi, line 64: You don't need to check for hover state, the browser does that automatically

packages/runtime/src/widgets/Row.tsx
```diff
@@ -26,11 +28,19 @@ export const Row: React.FC<FlexboxProps> = (props) => {
   const renderedChildren = useMemo(() => {
     return children ? EnsembleRuntime.render(children) : null;
   }, [children]);
-
+  const action = useEnsembleAction(props?.onTap);
```
evshi, line 31: Destructure this from the props

evshi (commented):
Can you also add this to Column since it's part of the spec

packages/runtime/src/widgets/Row.tsx
sibteali786: got it @evshi, but i have one question and it is solely for reason of curiosity. If you can answer it please. 
Is there any performance or code quality reason to destructure first and then use vs directly using a prop.

evshi: Yes, in this case if you destructure it first, then it won't be passed into the register bindings so saves extra processing

apps/kitchen-sink/src/ensemble/screens/home.yaml
```diff
@@ -14,6 +14,20 @@ View:
...
+            children:
+              - Text:
+                  text: Testing the Row 
+        - Column:
+            onTap:
+              executeCode:  |
+                console.log('Column Clicked');
+            children:
+              - Text:
+                  text: Testing the Column
```
evshi, line 30: Move this to either actions or widgets screen.

evshi (approved):
LGTM but looks like you have linter errors

packages/runtime/src/widgets/Row.tsx
```diff
@@ -51,6 +59,7 @@ export const Row: React.FC<FlexboxProps> = (props) => {
         flexDirection: "row",
         flexFlow: "unset",
         flexGrow: "unset",
+        cursor: "pointer",
```
evshi, line 62: This should only be pointer if the action callback exists. Same with Column. Can you also update the GridRow with same behavior?

anserwaseem:
@sibteali786 can you also add `visible` property in Row and Column as well?

sibteali786:
> @sibteali786 can you also add `visible` property in Row and Column as well?

jo hukam sarkar

evshi (approved):

PR #319, opened 2024-01-16
added support of item-template inside the dropdown

### LINKED ISSUE #250: Dropdown should support Item-template and label should also accpet a widget

**Additional context**

sagardspeed2:
## Describe your changes
Need to add following support inside the dropdown
1. item template
2. label should support widget

Closes #
closes #250 

- [ ] I have added tests


packages/runtime/src/widgets/Dropdown.tsx
```diff
@@ -39,6 +63,41 @@ const Dropdown: React.FC<DropdownProps> = (props) => {
...
+    }
+
+    if (isObject(itemTemplate)) {
+      return namedData.map((item: unknown, index) => {
+        return (
+          <Select.Option
+            key={index}
+            value={evaluate(defaultScreenContext, itemTemplate.value, {
+              [itemTemplate.name]: get(item, itemTemplate.name) as unknown,
+            })}
```
sagardspeed2, line 90: @evshi I need your help here,

as right now I am getting the error of 
```Unsafe assignment of an `any` value```

and If I wrap this with either Number or String, there is no error.
but we can't be sure that every time we will get Number or String as a value.

Comment (evshi) [reply packages/runtime/src/widgets/Dropdown.tsx:90]: `evaluate` takes a type argument that will case the return value

File: .changeset/tender-plums-crash.md [hunk 2]
```diff
@@ -0,0 +1,6 @@
+---
+"@ensembleui/react-framework": patch
+"@ensembleui/react-runtime": patch
+---
+
+Fix the crashing of charts
```
evshi, line 6: Is this the right changeset?

apps/kitchen-sink/src/ensemble/screens/widgets.yaml
```diff
@@ -54,10 +54,37 @@ View:
...
+                  item-template:
+                    data:
+                      [
+                        { name: "hello", value: 11 },
+                        { name: "buyyy", value: 22 },
+                      ]
+                    name: product
+                    value: ${product.value}
+                    template:
+                      Text:
```
evshi, line 68: Does this follow the flutter sides contract? I think we also need to bind the value

packages/runtime/src/widgets/Dropdown.tsx
```diff
@@ -1,34 +1,59 @@
...
 
 export type DropdownProps = {
   label?: string;
   hintText?: string;
   value?: Expression<string | number>;
-  items: SelectOption[];
+  items?: SelectOption[];
   onItemSelect: EnsembleAction;
   autoComplete: Expression<boolean>;
+  "item-template"?: DropdownItemTemplace;
```
evshi, line 38: There is a shared type for this

packages/runtime/src/widgets/Dropdown.tsx
```diff
@@ -39,6 +64,43 @@ const Dropdown: React.FC<DropdownProps> = (props) => {
...
     [action],
   );
+
+  const { namedData } = useTemplateData({
+    data: itemTemplate?.data,
+    name: itemTemplate?.name,
+  });
+
+  const options = useMemo(() => {
+    if (values?.items) {
```
evshi, line 74: Can you confirm what the behavior is for flutter? What is the order of preference or perhaps they concatenate the options if both exist?

packages/runtime/src/widgets/Dropdown.tsx
sagardspeed2: Yup, I know. but here we need extra args inside the `item-template`. so I used it like this

apps/kitchen-sink/src/ensemble/screens/widgets.yaml
sagardspeed2: Yes, I have seen the definition of flutter implementation for the item-template inside the dropdown. they are using this way to passing and evaluate the value on each option

packages/runtime/src/widgets/Dropdown.tsx
sagardspeed2: It concatenate both, items and item-template if available, and priority is given to items. so as per that I updated the PR

packages/runtime/src/widgets/Dropdown.tsx
evshi: You can use type composition: `HasItemTemplate & { value: Expression<string> }`

evshi (commented):
Looks good, just fix changeset and item template typing

evshi (approved):

PR #320, opened 2024-01-17
added navigateUrl API

### LINKED ISSUE #314: implement useNavigateURL action

**Additional context**

sagardspeed2:
## Describe your changes
Added new api for navigate screen using url - `navigateUrl`

Closes #
closes #314 

- [ ] I have added tests

### DIFF
--- a/apps/kitchen-sink/src/ensemble/screens/actions.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/actions.yaml
@@ -19,9 +19,20 @@ View:
               names: heading-1
             text: ensemble.navigateScreen()
         - Markdown:
-            text: use `ensemble.navigateScreen(screenname, data)` for navigate screen programmatically. click below button to check.
+            text: use `ensemble.navigateScreen({screenname, data})` for navigate screen programmatically. click below button to check.
         - Button:
             label: navigate screen
             onTap:
               executeCode: |
                 ensemble.navigateScreen({name: "home", inputs: {hello: colors.primary['900']}})
+        - Text:
+            styles:
+              names: heading-1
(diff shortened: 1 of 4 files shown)


apps/kitchen-sink/src/ensemble/screens/actions.yaml
```diff
@@ -19,9 +19,20 @@ View:
...
+        - Text:
+            styles:
+              names: heading-1
+            text: ensemble.navigateUrl()
+        - Markdown:
+            text: use `ensemble.navigateUrl(path, data)` for navigate screen programmatically. click below button to check.
+        - Button:
+            label: navigate url
+            onTap:
+              executeCode: |
```
evshi, line 37: This should also be usable as a YAML construct. You can follow the useNavigateScreen as an example

sagardspeed2: okay, let me work on that also

evshi (dismissed):

sagardspeed2:
If the path start with `/` then it will considered as start of path which will resolved with the domain name, and if path is not start with `/` then it will considered as a relative path and resolved with the current directory.

this is already added in this PR.

sibteali786:
Yes i agree @sagardspeed2 but the issue arises when we are changing role-based paths like from advocate/home even if we use /manager/dashboard it only concatenates the URL afterwards like this /advocate/manager/dasboard, which is not desired output.

- I suggest using window.href.location to change URL when we are changing roles specially, because otherwise using useLocation hook its not possible to be done as per my knowledge.
@evshi @sagardspeed2

evshi (approved):

PR #321, opened 2024-01-17
support returning last statement of multiline js

### LINKED ISSUE #311: autosave prettier causes error when object is a but larger in ScriptFile
**Describe the bug**
ScriptFile causes error in evaluator ( evaluate.ts ) when an object span multiple lines 

**To Reproduce**
We can use this snippet in the scriptFile ( replace screen name with any screen name ) 
```js
ensemble.navigateScreen({
  name: 'PostEnrollment',
  inputs: { patientMrn: '2138', program: '', isPostEnroll: false },
});
```
It will produce this error 
Screenshot 2024-01-14 at 2 27 05 PM
otherwise if similar is used but being on similar line it works fine 
```js
ensemble.navigateScreen({ name: 'PostEnrollment', inputs: { patientMrn: '2138', program: '', isPostEnroll: false }});
```

**Expected behavior**

**Screenshots**

**Additional context**

anserwaseem:

Closes #311

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/evaluate.ts
+++ b/packages/framework/src/evaluate.ts
@@ -49,22 +49,9 @@ const formatJs = (js?: string): string => {
 
   // multiline js
   if (sanitizedJs.includes("\n")) {
-    const lines = sanitizedJs.split("\n");
-
-    let lineIndex = -1;
-    for (let i = lines.length - 1; i >= 0; i--) {
-      if (lines[i].includes("(")) {
-        lineIndex = i;
-        break;
-      }
-    }
-    if (lineIndex !== -1) {
-      lines[lineIndex] = `return ${lines[lineIndex]}`;
-    }
-
     return `
       return (function() {
-         ${lines.join("\n")}
+         ${sanitizedJs}
       }())
     `;
   }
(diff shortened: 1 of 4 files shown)


packages/framework/src/evaluate.ts
```diff
@@ -50,8 +50,17 @@ const formatJs = (js?: string): string => {
   // multiline js
   if (sanitizedJs.includes("\n")) {
     const lines = sanitizedJs.split("\n");
-    const lastLine = last(lines);
-    lines.splice(lines.length - 1, 1, `return ${String(lastLine)}`);
+
+    let lineIndex = -1;
+    for (let i = lines.length - 1; i >= 0; i--) {
+      if (lines[i].includes("(")) {
```
evshi, line 56: This isn't going to work if there are two function calls in the same block. I think we just remove the return insertion for multiline blocks and require an explicit return value if needed. I'm pretty sure the only place where the return is required is the chart config so that will need to be updated.

evshi (commented):
@anserwaseem does this break charts?

anserwaseem:
> @anserwaseem does this break charts?

no
image

evshi (approved):

PR #325, opened 2024-01-19
added collapsible widget

### LINKED ISSUE #297: Implement Collapsible widget
It's a common UX pattern to have a section with an interactive header that can collapse the field(s) below the header.

Design and implement a Collapsible widget that allows user to specify content and header widgets, and state/mutators for whether it is/is not collapsed.

**Additional context**

sagardspeed2:
## Describe your changes
Added a collapsible widget with functionality of accordion.
Features:
Item template
can pass custom icons
can change header and content background color

Closes #
closes #297 

- [ ] I have added tests

### DIFF
--- a/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
@@ -20,6 +20,29 @@ View:
             text: Widgets
         - Markdown:
             text: More to come! In the meantime, checkout the Ensemble [documentation](https://docs.ensembleui.com/).
+
+        - Markdown:
+            text: "## Collapse"
+        - Collapsible:
+            id: collapseWidget
+            value:
+              - "1"
+            expandIconPosition: "end"
+            items:
+              - key: "1"
+                label:
+                  Text:
+                    text: Here is my label 1
+                children: hiii
+              - key: "2"
+                label: Here is my label 2
+                children:
+                  Text:
+                    text: Here is my value 2
+            onCollapse:
+              executeCode: |
+                console.log('Hello', data, collapseWidget)
+
         - Card:
             styles:
               maxWidth: unset
--- a/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
@@ -25,8 +25,10 @@ View:
             text: "## Collapse"
         - Collapsible:
             id: collapseWidget
-            value:
-              - "1"
+            collpaseIcon:
+              name: HomeOutlined
+            expandIcon:
+              name: WidgetsOutlined
             expandIconPosition: "end"
             items:
               - key: "1"
--- a/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
@@ -30,6 +30,24 @@ View:
             expandIcon:
               name: WidgetsOutlined
             expandIconPosition: "end"
+            item-template:
+              data:
+                [
+                  { key: "1", name: "Label 1", std: "Hello" },
+                  { key: "2", name: "Label 2", std: "Hello 123" },
+                  { key: "3", name: "Label 3", std: "Hello 456" },
+                ]
+              name: item
+              template:
+                collapsibleItem:
+                  key: ${item.key}
+                  label: ${item.name}
+                  children: ${item.std}
+            onCollapse:
+              executeCode: |
+                console.log('Hello', data, collapseWidget)
+        - Collapsible:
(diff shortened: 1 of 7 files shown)


apps/kitchen-sink/src/ensemble/screens/widgets.yaml
```diff
@@ -20,6 +20,29 @@ View:
...
             text: More to come! In the meantime, checkout the Ensemble [documentation](https://docs.ensembleui.com/).
+
+        - Markdown:
+            text: "## Collapse"
+        - Collapsible:
+            id: collapseWidget
+            value:
+              - "1"
+            expandIconPosition: "end"
+            items:
```
evshi, line 31: This should take a generic list of children like Column/Row

apps/kitchen-sink/src/ensemble/screens/widgets.yaml
```diff
@@ -20,6 +20,29 @@ View:
...
         - Markdown:
             text: More to come! In the meantime, checkout the Ensemble [documentation](https://docs.ensembleui.com/).
+
+        - Markdown:
+            text: "## Collapse"
+        - Collapsible:
+            id: collapseWidget
+            value:
+              - "1"
+            expandIconPosition: "end"
```
evshi, line 30: How about supporting either a header widget and allow them to customize the header entirely OR support a title and icon

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -0,0 +1,84 @@
...
+export type CollapsibleProps = {
+  items: CollapsibleItem[];
+  expandIconPosition?: "start" | "end";
+  value: Expression<string>[];
+  onCollapse?: EnsembleAction;
+  isAccordion?: boolean;
+} & EnsembleWidgetProps;
+
+export const Collapsible: React.FC<CollapsibleProps> = (props) => {
+  const [activeValue, setActiveValue] = useState<string[]>(props.value);
```
evshi, line 31: Does this represent whether it's collapsed or not?

apps/kitchen-sink/src/ensemble/screens/widgets.yaml
sagardspeed2: I already added a support for that, user can pass any widget as a header for each collapsible item.
for example,
```js
items:
  - key: "1"
    label:
      Text:
        text: Here is my label 1
    children: hiii
  - key: "2"
    label: Here is my label 2
    children:
      Text:
        text: Here is my value 2
```

here label represent the header of collapsible and children represent the content which will be shown when the item will be collapse

packages/runtime/src/widgets/Collapsible.tsx
sagardspeed2: This collapsible widget takes an array of labels and children as input. Based on this array, it creates a list of collapsibles. If the activeValue state has any value, it indicates that one or multiple collapsibles from that array are collapsed.

apps/kitchen-sink/src/ensemble/screens/widgets.yaml
sagardspeed2: @evshi I think here we do not need of passing children like column/rows,
because here we used prebuilt collapsible widget of antd design.

here is the reference,
https://ant.design/components/collapse

packages/runtime/src/widgets/Collapsible.tsx
evshi: I see. We will need to collapse items individually and read their value. The invokable method that is registered should be a callback that takes the index and value to make it easier to use `(index: int, value: string) => void` over passing an entire array.

evshi (commented):
@anserwaseem can you please review and see if it meets the requirements you need?

anserwaseem:
> @anserwaseem can you please review and see if it meets the requirements you need?

@sagardspeed2 would need few more things:
1. `item-template` 
2. ability to give my own icon
3. styles for collapsible head (I need to change its background colour) - low priority

evshi:
@sagardspeed2 something like this?

```yaml
Collapsibe:
  item-template:
     data
     name: item
     template:
         CollapsibleItem:
             label ${item.title}
             key: ${item.id}
             item:
                Text:
                    text: ${item.value}
```

sagardspeed2:
> > @anserwaseem can you please review and see if it meets the requirements you need?
> 
> @sagardspeed2 would need few more things:
> 
> 1. `item-template`
> 2. ability to give my own icon
> 3. styles for collapsible head (I need to change its background colour) - low priority

Added item-template, ability to give custom icon, also added ability to change the collapsible header and content background color

evshi (approved):
GEFN to unblock @anserwaseem. @sagardspeed2 please add a changeset

PR #328, opened 2024-01-22
Data grid row selection

sibteali786:
## Describe your changes
- added ability for users to select a particular row or all rows 
- exposed onRowSelect action
### Screenshots [Optional]

https://github.com/EnsembleUI/ensemble-react/assets/62849614/c361fb73-c452-40a6-8a5b-a44c7092f93d

Closes #

- [ ] I have added tests

### DIFF
--- a/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
@@ -301,6 +301,9 @@ View:
               name: dummyData
               template:
                 DataRow:
+                  onRowsSelected:
+                    executeCode:  |
+                      console.log("data", selectedRows, selectedRowKeys);
                   children:
                     - Text:
                         id: ${dummyData.patient+"Name"}
--- a/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
@@ -277,7 +277,37 @@ View:
                                   ]
                               }
                           }
-
+        - Markdown:
+            text: "## DataGrids"
+        - DataGrid:
+            id: "myDataGridExample"
+            allowSelection: true
+            styles:
+              headerStyle:
+                backgroundColor: "white"
+                fontSize: "12px"
+                textColor: "#747184"
+                hasDivider: true
+                borderBottom: false
+            DataColumns:
+              - label: PATIENT
+              - label: ADVOCATE
+            item-template:
+              data:
+                [
+                  { key: "1", patient: "John", advocate: "Mike Ross" },
+                  { key: "2", patient: "Reuben", advocate: "Harvey Dent" },
+                ]
+              name: dummyData
+              template:
+                DataRow:
+                  children:
+                    - Text:
+                        id: ${dummyData.patient+"Name"}
+                        text: ${dummyData.patient}
+                    - Text:
+                        id: ${dummyData.patient+"Adv"}
+                        text: ${dummyData.advocate}
         - Card:
             styles:
               maxWidth: unset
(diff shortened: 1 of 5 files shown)


apps/kitchen-sink/src/ensemble/screens/widgets.yaml
```diff
@@ -217,7 +217,40 @@ View:
...
+                [
+                  { key: "1", patient: "John", advocate: "Mike Ross" },
+                  { key: "2", patient: "Reuben", advocate: "Harvey Dent" },
+                ]
+              name: dummyData
+              template:
+                DataRow:
+                  onRowsSelected:
+                    executeCode:  |
+                      console.log("data", selectedRows, selectedRowKeys);
```
sagardspeed2, line 246: @sibteali786  Can we move this part outside of the item-template or template ?

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -88,12 +107,29 @@ export const DataGrid: React.FC<GridProps> = (props) => {
...
           return { onClick: () => onTapActionCallback(record, recordIndex) };
         }}
-        dataSource={namedData}
-        key={resolvedWidgetId}
         ref={rootRef}
+        rowKey={(data: unknown) => {
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+          const x = get(data, itemTemplate.name);
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-member-access
+          return x.key ?? "";
```
sagardspeed2, line 120: @sibteali786 Can we make this as dynamic item, as each time there will be not a key, sometime user can use id or any other field also

apps/kitchen-sink/src/ensemble/screens/widgets.yaml
sibteali786: makes sense, let me do that

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
sibteali786: agree just on it.

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -47,10 +47,13 @@ export interface DataGridStyles extends Partial<EnsembleWidgetStyles> {
 }
 
 export type GridProps = {
+  allowSelection?: boolean;
+  onRowsSelected?: EnsembleAction;
   DataColumns: DataColumn[];
   "item-template": {
     data: Expression<object>;
     name: string;
+    uniqueIdentifier?: Expression<string>;
```
evshi, line 56: This is coming up more and more....Let's call this `key`

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -64,16 +67,32 @@ export interface DataGridRowTemplate {
 }
 
 export const DataGrid: React.FC<GridProps> = (props) => {
+  const [selectedRow, setSelectedRow] = useState<object[]>();
```
evshi, line 70: Plural selectedRows

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -88,12 +107,33 @@ export const DataGrid: React.FC<GridProps> = (props) => {
...
         ref={rootRef}
+        rowKey={(data: unknown) => {
+          const identifier: string = evaluate(
+            defaultScreenContext,
+            itemTemplate.uniqueIdentifier,
+            {
+              [itemTemplate.name]: get(data, itemTemplate.name) as unknown,
+            },
+          );
+          return identifier ?? "";
```
evshi, line 124: Maybe as a fallback we do a hash of the data

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
sibteali786: since its already being used below in rowSelection's onChange so i am renaming to rowsSelected

evshi (approved):

PR #329, opened 2024-01-22
add styles enhancements in various widgets

anserwaseem:

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/body.tsx
+++ b/packages/runtime/src/runtime/body.tsx
@@ -51,7 +51,7 @@ export const EnsembleBody: React.FC<EnsembleBodyProps> = ({
     styles: {
       height:
         // We only need to do this calculation if there's a footer otherwise, page can flow freely
-        isModal && footer
+        !isModal && footer
           ? `calc(100vh - ${marginTop} - ${marginBottom})`
           : undefined,
       overflow: "auto",
--- a/packages/runtime/src/runtime/body.tsx
+++ b/packages/runtime/src/runtime/body.tsx
@@ -51,7 +51,7 @@ export const EnsembleBody: React.FC<EnsembleBodyProps> = ({
     styles: {
       height:
         // We only need to do this calculation if there's a footer otherwise, page can flow freely
-        !isModal && footer
+        isModal && footer
           ? `calc(100vh - ${marginTop} - ${marginBottom})`
           : undefined,
       overflow: "auto",
--- a/packages/runtime/src/widgets/Card.tsx
+++ b/packages/runtime/src/widgets/Card.tsx
@@ -52,7 +52,8 @@ export const Card: React.FC<CardProps> = ({ children, styles }) => {
     <div
       className={styles?.names}
       style={{
-        display: "grid",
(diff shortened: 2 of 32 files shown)


sibteali786 (approved):
thanks for adding styles to dropdown, a much awaited thing @anserwaseem

packages/runtime/src/widgets/Card.tsx
```diff
@@ -41,17 +42,37 @@ export const Card: React.FC<CardProps> = ({ children, styles }) => {
...
+            ${
+              gap
+                ? `
+                    display: block !important;
+                    margin-top: ${gap} !important;
+                  `
+                : ""
+            }
+          }
+        `}
```
evshi, line 64: Why is this needed?

packages/runtime/src/widgets/Dropdown.tsx
```diff
@@ -102,25 +129,83 @@ const Dropdown: React.FC<DropdownProps> = (props) => {
...
+          marginBottom: "0px",
+          ...(values?.styles?.visible === false
+            ? { display: "none" }
+            : undefined),
+          ...values?.labelStyle,
+        }}
+      >
+        {values?.id ? (
+          <style>{`
+        [data-testid=${values?.id}] .ant-select .ant-select-selector {
```
evshi, line 147: Instead of using the testId, use the stylenames.

packages/runtime/src/runtime/body.tsx
```diff
@@ -51,7 +51,7 @@ export const EnsembleBody: React.FC<EnsembleBodyProps> = ({
     styles: {
       height:
         // We only need to do this calculation if there's a footer otherwise, page can flow freely
-        !isModal && footer
+        isModal && footer
```
evshi, line 54: This would break the pages where there is a footer, like the Workflow screen

packages/runtime/src/widgets/Card.tsx
evshi: In that case, why not just allow setting the styles on the card element directly?

packages/runtime/src/widgets/Card.tsx
```diff
@@ -41,13 +44,17 @@ export const Card: React.FC<CardProps> = ({ children, styles }) => {
...
+
+  const mergedStyles = { ...defaultStyles, ...styles };
   const { shadowOffset, shadowBlur, shadowSpread, shadowColor } = mergedStyles;
+
   return (
     <div
+      className={styles?.names}
       style={{
-        ...mergedStyles,
+        display: "grid",
```
evshi, line 55: Why display: grid here?

evshi (approved):

packages/runtime/src/widgets/Card.tsx
anserwaseem: cause default block display isn't letting `gap` property have any effect

evshi: We can use display: flex

evshi (approved):

PR #342, opened 2024-01-24
added widget support in form items

### LINKED ISSUE #338: Support widget for form item label and checkbox trailing/leading text

sagardspeed2:
## Describe your changes
Added widget support feature in form items

Closes #
closes #338 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Form/Checkbox.tsx
+++ b/packages/runtime/src/widgets/Form/Checkbox.tsx
@@ -46,7 +46,6 @@ export const CheckboxWidget: React.FC<CheckBoxProps> = (props) => {
 
   return (
     <EnsembleFormItem valuePropName="checked" values={values}>
-      {leadingContent}
       <Checkbox
         checked={Boolean(checked)}
         disabled={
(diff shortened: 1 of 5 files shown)


evshi (commented):
@sagardspeed2 per the ticket can you also add widget support for the label attribute?

packages/runtime/src/widgets/Form/Checkbox.tsx
```diff
@@ -36,6 +46,7 @@ export const CheckboxWidget: React.FC<CheckBoxProps> = (props) => {
 
   return (
     <EnsembleFormItem valuePropName="checked" values={values}>
+      {leadingContent}
```
evshi, line 49: Does this break the form submission?

sagardspeed2: Yup, when we used the leadingContent, the checkbox checked stat is not updating.
so if we are not using the leadingContent, should I remove the leadingText prop from the definition ?

packages/runtime/src/widgets/Form/Checkbox.tsx
```diff
@@ -37,7 +50,7 @@ export const CheckboxWidget: React.FC<CheckBoxProps> = (props) => {
             : undefined),
         }}
       >
-        {values?.trailingText}
+        {trailingContent}
```
evshi, line 53: Trailing and leading text are mutually exclusive so we can support either or and use direction: LTR/RTL to make it appear before and after

evshi (approved):

PR #365, opened 2024-01-28
Stylistic-updates

sibteali786:
## Describe your changes
- added cursor pointer for Icon 
- added divider in Popupmenu 
- added group selections for Dropdown

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/customWidget.tsx
+++ b/packages/runtime/src/runtime/customWidget.tsx
@@ -13,7 +13,6 @@ import { useEnsembleAction } from "./hooks/useEnsembleAction";
 
 export interface CustomWidgetProps {
   inputs: Record<string, unknown>;
-  onLoad: EnsembleAction;
 }
 
 export const createCustomWidget = (
--- a/packages/runtime/src/runtime/customWidget.tsx
+++ b/packages/runtime/src/runtime/customWidget.tsx
@@ -13,6 +13,7 @@ import { useEnsembleAction } from "./hooks/useEnsembleAction";
 
 export interface CustomWidgetProps {
   inputs: Record<string, unknown>;
+  onLoad: EnsembleAction;
 }
 
 export const createCustomWidget = (
--- a/packages/runtime/src/runtime/customWidget.tsx
+++ b/packages/runtime/src/runtime/customWidget.tsx
@@ -25,7 +25,7 @@ export const createCustomWidget = (
     });
     return (
       <CustomScopeProvider value={values ?? inputs}>
-        <OnLoadAction action={widget.onLoad} context={values ?? inputs}>
+        <OnLoadAction action={widget?.onLoad} context={values ?? inputs}>
           {EnsembleRuntime.render([widget.body])}
         </OnLoadAction>
       </CustomScopeProvider>
--- a/packages/runtime/src/widgets/Column.tsx
+++ b/packages/runtime/src/widgets/Column.tsx
@@ -36,7 +36,7 @@ export const Column: React.FC<FlexboxProps> = (props) => {
     if (!action) {
       return;
     }
-    action.callback();
+    action.callback({ ...parentScope });
   }, [action]);
   return (
     <Col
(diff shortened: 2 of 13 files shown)


packages/runtime/src/runtime/customWidget.tsx
```diff
@@ -19,9 +25,35 @@ export const createCustomWidget = (
     });
     return (
       <CustomScopeProvider value={values ?? inputs}>
-        {EnsembleRuntime.render([widget.body])}
+        <OnLoadAction action={widget?.onLoad} context={{ ...inputs }}>
```
evshi, line 28: Should be values and then inputs just like line 21

packages/runtime/src/runtime/customWidget.tsx
```diff
@@ -13,6 +13,7 @@ import { useEnsembleAction } from "./hooks/useEnsembleAction";
 
 export interface CustomWidgetProps {
   inputs: Record<string, unknown>;
+  onLoad: EnsembleAction;
```
evshi, line 16: The props are not needed if it's on the widget model.

packages/runtime/src/widgets/Column.tsx
```diff
@@ -36,7 +36,7 @@ export const Column: React.FC<FlexboxProps> = (props) => {
     if (!action) {
       return;
     }
-    action.callback();
+    action.callback({ ...parentScope });
```
evshi, line 39: The scope should be injected by the action, not passed in.

packages/runtime/src/widgets/Icon.tsx
```diff
@@ -30,7 +33,7 @@ export const Icon: React.FC<IconProps> = ({ onTap, ...props }) => {
   return (
     <IconComponent
       className={values?.styles?.names}
-      onClick={(): unknown => onTapActionCallback?.callback()}
+      onClick={(): unknown => onTapActionCallback?.callback({ ...scope })}
```
evshi, line 36: Scope should be injected by the action, not passed in. Which action is missing the scope?

packages/runtime/src/widgets/Row.tsx
```diff
@@ -34,7 +34,7 @@ export const Row: React.FC<FlexboxProps> = (props) => {
     if (!action) {
       return;
     }
-    action.callback();
+    action.callback({ ...parentScope });
```
evshi, line 37: Same thing, scope should be injected by the action. The parent scope should get removed here, it's automatically provided in the custom scope provider.

packages/runtime/src/widgets/Skeleton.tsx
```diff
@@ -16,7 +17,7 @@ export const Skeleton: React.FC<SkeletonProps> = (props) => {
       animation={values?.useShimmer ? "wave" : false}
       height={values?.styles?.height}
       style={values?.styles}
-      variant="rectangular"
+      variant={values?.variant ? values?.variant : "rectangular"}
```
evshi, line 20: Can use nullish coalescing ??

packages/runtime/src/widgets/Icon.tsx
evshi: It's executeCode that is not injecting the custom scope, not the onTap trigger. This should be fixed on the latest main.

packages/runtime/src/widgets/ToggleButton.tsx
```diff
@@ -30,6 +30,7 @@ interface ToggleButtonPropsStyles {
   borderWidth?: number;
   selectedBorderColor?: Expression<string>;
   selectedBorderWidth?: number;
+  buttonWidth?: Expression<string>;
```
evshi, line 33: What are you using this for? The button widths can derive their width from the overall Toggle Button width

sibteali786: well it is not, in my case i wanted to apply 50 % width to two toggle buttons

evshi (approved):

PR #366, opened 2024-01-29
fix: added stickey header and footer in data grid

### LINKED ISSUE #364: Enable sticky header and footer in DataGrid

sagardspeed2:
## Describe your changes
Added feature for enable stickey header and footer in data grid widget

Closes #
closes #364 

- [ ] I have added tests

### DIFF
--- a/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
@@ -49,8 +49,8 @@ View:
                     - label: ADVOCATE
                     - label: ICON
                     - label: MENU
-                  scrollable: true
-                  scrollHeight: 400
+                  scroll:
+                    scrollHeight: 400
                   item-template:
                     data:
                       [
--- a/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
@@ -49,200 +49,12 @@ View:
                     - label: ADVOCATE
                     - label: ICON
                     - label: MENU
-                  scroll:
-                    scrollHeight: 400
                   item-template:
                     data:
                       [
                         { key: "1", patient: "John", advocate: "Mike Ross" },
                         {
                           key: "2",
-                          patient: "Reuben",
-                          advocate: "Harvey Dent",
-                        },
-                        { key: "3", patient: "John", advocate: "Mike Ross" },
-                        {
-                          key: "4",
-                          patient: "Reuben",
-                          advocate: "Harvey Dent",
-                        },
-                        { key: "5", patient: "John", advocate: "Mike Ross" },
-                        {
-                          key: "6",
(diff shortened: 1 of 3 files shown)


packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -168,14 +171,16 @@ export const DataGrid: React.FC<GridProps> = (props) => {
               }
             : undefined
         }
+        scroll={
+          values?.scrollable ? { y: values.scrollHeight || 150 } : undefined
+        }
```
evshi, line 176: Having two properties seems redundant here. Let's just expose the scroll config object

apps/kitchen-sink/src/ensemble/screens/widgets.yaml
```diff
@@ -49,11 +49,203 @@ View:
...
                     - label: ICON
                     - label: MENU
+                  scroll:
+                    scrollHeight: 400
                   item-template:
                     data:
                       [
                         { key: "1", patient: "John", advocate: "Mike Ross" },
-                        { key: "2", patient: "Reuben", advocate: "Harvey Dent" },
+                        {
```
evshi, line 58: Bind to endpoint instead of static data

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -55,17 +55,24 @@ export interface DataGridRowTemplate {
...
 
+export interface DataGridScrollable {
+  scrollHeight: string;
+}
+
 export type GridProps = {
   allowSelection?: boolean;
   onRowsSelected?: EnsembleAction;
   DataColumns: DataColumn[];
+  scrollable?: boolean;
```
evshi, line 66: Remove unused properties

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -55,17 +55,24 @@ export interface DataGridRowTemplate {
   };
 }
 
+export interface DataGridScrollable {
+  scrollHeight: string;
```
evshi, line 59: Let's add width while you're at it

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -142,6 +149,7 @@ export const DataGrid: React.FC<GridProps> = (props) => {
         onRow={(record, recordIndex) => {
           return { onClick: () => onTapActionCallback(record, recordIndex) };
         }}
+        pagination={values?.hidePagination ? false : undefined}
```
evshi, line 152: `values.hidePagination !== true`

sagardspeed2: We can't update this condition that return the bool, as pagination only accept either false or undefined.

evshi (approved):

PR #369, opened 2024-01-29
OnTap-updates

sibteali786:
## Describe your changes
- added onTap to have access to local scope 
- similar for pickFiles action

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
+++ b/packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
@@ -331,17 +331,22 @@ export const usePickFiles: EnsembleActionHook<PickFilesAction> = (
         setFiles(selectedFiles);
       }
     };
+
+    return input;
+  }, [action?.allowMultiple, action?.allowedExtensions]);
+  const scope = useCustomScope();
+  useEffect(() => {
     return () => {
       inputEl.remove();
     };
   }, [inputEl, files]);
 
   useEffect(() => {
-    if (!isEmpty(values?.files) && isComplete === false) {
-      onCompleteAction?.callback({ files: values?.files });
+    if (!isEmpty(values?.files) && !isComplete) {
+      onCompleteAction?.callback({ files: values?.files, ...scope });
       setIsComplete(true);
     }
-  }, [values?.files, onCompleteAction, isComplete]);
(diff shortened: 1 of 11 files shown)


packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -323,10 +324,10 @@ export const usePickFiles: EnsembleActionHook<PickFilesAction> = (
 
   useEffect(() => {
     if (!isEmpty(values?.files) && !isComplete) {
-      onCompleteAction?.callback({ files: values?.files });
+      onCompleteAction?.callback({ files: values?.files, ...scope });
```
evshi, line 327: What action are you trying to invoke here that is missing the scope?

sibteali786: i am trying to access the item-template `name` under `pickFiles` to modify the the data

evshi: Can you share the YAML block? The `pickFiles` action should have access to the custom scope via `useRegisterBindings`

sibteali786: yes here 
```yaml
Button:
                              label: Upload File
                              onTap:
                                pickFiles:
                                  id: ${"uploadDoc"+docList.index}
                                  allowMultiple: true
                                  allowedExtensions:
                                    - pdf
                                    - docx
                                  onComplete:
                                    executeCode: |
                                      const fileName = files[0].name;
                                      const date = formatDate(files[0].lastModifiedDate);
                                      const data = ensemble.storage.get('docsData');
                                      const newData = data.map((d) => {
                                        if(d.index === docList.index){
                                          d.fileName = fileName;
                                          d.date = date;
                                        }
                                        return d;
                                      });
                                      ensemble.storage.set('docsData', newData);
```

evshi: So the ID is the issue? For executeCode, we would need to add the custom scope hook there.

sibteali786: so are we gonna approve it @evshi ?

PR #370, opened 2024-01-29
hotfix/add-default-selection-dropdown

sagardspeed2:
## Describe your changes
Added missing functionality of selecting default value in dropdown

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Form/Dropdown.tsx
+++ b/packages/runtime/src/widgets/Form/Dropdown.tsx
@@ -131,7 +131,7 @@ const Dropdown: React.FC<DropdownProps> = (props) => {
     }
 
     return dropdownOptions;
-  }, [values?.items, namedData, itemTemplate]);
+  }, [values?.items, values?.id, namedData, itemTemplate]);
 
   const { backgroundColor: _, ...formItemStyles } = values?.styles ?? {};
 


anserwaseem (changes requested):
I think we need `defaultValue={values?.selectedValue}` on antd Select component, don't we?

packages/runtime/src/widgets/Form/Dropdown.tsx
```diff
@@ -131,7 +131,7 @@ const Dropdown: React.FC<DropdownProps> = (props) => {
     }
 
     return dropdownOptions;
-  }, [values?.items, namedData, itemTemplate]);
+  }, [values?.items, values?.id, namedData, itemTemplate]);
```
evshi, line 134: How does this fix the issue? Seems like values.id is only used for classname?

evshi (commented):
@anserwaseem if you're having issues with default values in dropdowns outside of forms, do not use `id` property or replace `id` with `testId`. Adding the `id` makes the dropdown a controlled component and requires a form context.

anserwaseem:
@evshi yes, it works as you suggested. Thank you 🤩

sagardspeed2:
@evshi We need to add `defaultValue` props to `select` widget, 
because of when we remove the id, and use value with label , then label props doing conflict while setting the default value. so.

and if we use `defaultValue` prop, then if we are using id or not, it will not break any flow.

packages/runtime/src/widgets/Form/Dropdown.tsx
```diff
@@ -228,6 +228,7 @@ const Dropdown: React.FC<DropdownProps> = (props) => {
         <EnsembleFormItem values={values}>
           <Select
             className={`${values?.styles?.names || ""} ${id}_input`}
+            defaultValue={values?.selectedValue}
```
evshi, line 231: This should just be `values?.value`. `selectedValue` is fine for internal variable name but the form item interface always uses `value`

PR #379, opened 2024-01-31
dynamic datagrid columns

### LINKED ISSUE #371: Support dynamic DataGrid columns
DataGrid columns take a static list but sometimes we want to use the same DataGrid view for different datasets with variable columns.

- Add binding support for data column to be variable list
- Support key value on DataColumn
- Add item-template support for DataRow. The template should be applied over each column in the row.

**Additional context**

sagardspeed2:
## Describe your changes
Added support of dynamic datagrid columns

Closes #
closes #371 

- [ ] I have added tests


anserwaseem (approved):

apps/kitchen-sink/src/ensemble/screens/widgets.yaml
```diff
@@ -113,38 +124,8 @@ View:
                               id: ${dummyData.patient+"Name"}
                               text: ${dummyData.patient}
                           - Text:
-                              id: ${dummyData.patient+"Adv"}
+                              id: ${dummyData.advocate+"Adv"}
                               text: ${dummyData.advocate}
-                          - Icon:
```
evshi, line 118: Why are you removing these examples?

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -69,7 +73,7 @@ export interface DataGridScrollable {
 export type GridProps = {
   allowSelection?: boolean;
   onRowsSelected?: EnsembleAction;
-  DataColumns: DataColumn[];
+  DataColumns: DataColumn[] | Expression<object>;
```
evshi, line 76: Expression<DataColumn[] | string[]>

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -272,12 +277,14 @@ export const DataGrid: React.FC<GridProps> = (props) => {
               }
               render={(_: unknown, record: unknown): ReactElement => {
                 return (
-                  <DataCell
-                    columnIndex={index}
-                    data={record}
-                    scopeName={itemTemplate.name}
-                    template={itemTemplate.template}
-                  />
+                  <CustomScopeProvider value={record as CustomScope}>
```
evshi, line 280: You don't need this extra provider. The record is passed in as a prop to the data cell

packages/runtime/src/widgets/DataGrid/DataCell.tsx
```diff
@@ -10,14 +15,32 @@ export interface DataCellProps {
...
-    <CustomScopeProvider value={data as CustomScope}>
-      {EnsembleRuntime.render([cellTemplate])}
-    </CustomScopeProvider>
-  );
+  const { "item-template": itemTemplate, children } = template.properties;
+  const { namedData } = useTemplateData({
+    ...itemTemplate,
+  });
+
+  const cellData = useMemo(() => {
```
evshi, line 27: Let's get rid of this useMemo, I'm not sure there's a point here?

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
sagardspeed2: I think we need this one, because when we use the itemTemplate inside the DataRow, then in the DataCell, we are getting expression value and then we need to use `useTemplateData` here to convert this expression into the data. 

If I am not using this extra provider, then I am getting the error while `useTemplateData` that the variable is undefined.

packages/runtime/src/widgets/DataGrid/DataCell.tsx
sagardspeed2: There are two cases, either user pass the direct children array or the item template. and if user pass the item template then first we need to evaluate it's data expression into valid data object. and that's why I used here the memo.

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
evshi: Instead of injecting an extra scope provider, let's expose an extra `context` option on useTemplateData and pass in the record that way. This will avoid having a bunch of extra providers.

sagardspeed2: So basically when we pass the optional context to useTemplateData, then we need to merge that optional context with the useTemplateData's custom scope ?

evshi: we merge that context inside the evaluate hook, yes.

packages/framework/src/hooks/useTemplateData.ts
```diff
@@ -22,6 +23,7 @@ export interface TemplateDataProps {
 export const useTemplateData = ({
   data,
   name = "data",
+  context = {},
```
evshi, line 26: Make this optional instead of default arg

packages/runtime/src/widgets/DataGrid/DataCell.tsx
```diff
@@ -10,14 +14,29 @@ export interface DataCellProps {
...
+          {EnsembleRuntime.render([children[columnIndex]])}
+        </CustomScopeProvider>
+      ) : null}
+
+      {isObject(itemTemplate) && !isEmpty(namedData) && (
+        <CustomScopeProvider value={namedData[columnIndex] as CustomScope}>
+          {EnsembleRuntime.render([itemTemplate.template])}
+        </CustomScopeProvider>
+      )}
+    </>
```
evshi, line 40: Use separate if statements and have separate return statements. Easier to read

packages/framework/src/hooks/useTemplateData.ts
sagardspeed2: But if we make it optional then we will get context as undefined in case it is not passed in, and I think that will be making some issue while merging with the customScope.
is this right ?

packages/runtime/src/widgets/DataGrid/DataCell.tsx
```diff
@@ -10,14 +14,31 @@ export interface DataCellProps {
...
+
+  if (isObject(itemTemplate) && !isEmpty(namedData)) {
+    return (
+      <CustomScopeProvider value={namedData[columnIndex] as CustomScope}>
+        {EnsembleRuntime.render([itemTemplate.template])}
+      </CustomScopeProvider>
+    );
+  }
+
+  return <div />;
```
evshi, line 43: You don't need to return an element, or you can return null

evshi (approved):

evshi:
Please add changeset

PR #387, opened 2024-02-02
380 show dialog and close all dialogs api

### LINKED ISSUE #380: Implement showDialog Api
Implement `showDialog` api.

**Additional context**

anserwaseem:

Closes #380

- [ ] I have added tests


packages/runtime/src/runtime/showDialog.tsx
```diff
@@ -0,0 +1,73 @@
...
+  openModal?: (...args: any[]) => void;
+};
+
+export const showDialog = (props?: ShowDialogApiProps): void => {
+  const { action, openModal } = props ?? {};
+  if (!action || !openModal || !action?.widget) {
+    return;
+  }
+
+  const modalOptions = {
```
evshi, line 24: Can we make this consistent with the hook?

packages/runtime/src/runtime/showDialog.tsx
```diff
@@ -0,0 +1,73 @@
...
+  let inlineWidget: EnsembleWidget | undefined;
+  let customWidget: React.FC | ReactElement | undefined;
+
+  if (isString(widget?.name)) {
+    customWidget = WidgetRegistry.find(widget?.name);
+  } else if (widgetInputs !== undefined) {
+    customWidget = WidgetRegistry.find(widgetKey || "");
+  } else {
+    inlineWidget = cloneDeep(unwrapWidget(action?.widget));
+  }
```
evshi, line 53: Don't go looking up the component function, just create a widget model and pass it to EnsembleRuntime.

packages/runtime/src/runtime/showDialog.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const WidgetFn = customWidget as React.FC<unknown>;
+
+  const content = (
+    <CustomScopeProvider
+      value={{
+        ...(widgetInputs as CustomScope),
+        ...(widget?.inputs as CustomScope),
+      }}
+    >
```
evshi, line 63: Custom widget already handles this if the model is set up correctly.

packages/runtime/src/runtime/customWidget.tsx
```diff
@@ -39,20 +39,21 @@ const OnLoadAction: React.FC<
     context: Record<string, unknown>;
   }>
 > = ({ action, children, context }) => {
-  const onLoadAction = useEnsembleAction(action);
+  const onLoadActionRef = useRef(useEnsembleAction(action));
```
evshi, line 42: This is a sign you're breaking the render cycle in some way - why did you need this?

evshi (commented):
Keep the implementation as close to the hook as possible - the more you diverge, it's a sign you're not headed in the right direction. Stick to the contract of the widget model and EnsembleRuntime should do most of the heavy lifting.

packages/runtime/src/runtime/customWidget.tsx
anserwaseem: because custom widget's onLoad is firing twice :(

evshi: Like we discussed, if it's not obvious what the issue is, let's leave this implementation as is for now and do a follow up PR.

packages/runtime/src/runtime/showDialog.ts
```diff
@@ -0,0 +1,66 @@
...
+    return;
+  }
+
+  const widget = action?.widget;
+  const widgetKey = findKey(widget, isObject) || "";
+  let inlineWidget: EnsembleWidget | undefined;
+
+  if ((get(widget, widgetKey) as Record<string, unknown>)?.inputs) {
+    set(widget, "name", widgetKey);
+    set(widget, "properties", get(widget, widgetKey));
```
evshi, line 27: It's better to clone/create a new object here than modify the widget model inline. It's not clear to me why this requires special handling and can't be unwrapped with the util?

evshi (approved):

PR #397, opened 2024-02-07
convert storage to atom for ensemble user

### LINKED ISSUE #386: Allow functions on ensemble user atom
We need to be able to add functions on ensemble user atom so we can integrate with third party auth services. Functions cannot be stored in session/local storage and must be retained in memory.

Update the ensemble user atom to support function values. Attributes can still remain in session storage but the user atom should also be mirrored in the application context

**Additional context**

sagardspeed2:
## Describe your changes
to provide the function storage inside the ensemble user, I converted the storage atom to simple atom for ensemble user

Closes #386 

- [ ] I have added tests


packages/framework/src/hooks/useEnsembleUser.ts
```diff
@@ -0,0 +1,61 @@
+import { useAtom } from "jotai";
+import { atomWithStorage, createJSONStorage } from "jotai/utils";
+import { merge } from "lodash-es";
+import { useMemo } from "react";
+
+export type EnsembleUser = { accessToken?: string } & Record<string, unknown>;
+export type EnsembleUserBuffer = {
```
evshi, line 7: Don't expose this type externally. Just keep it all under EnsembleUser

evshi (commented):
Please fix typing and add changeset

packages/framework/src/evaluate/binding.ts
```diff
@@ -92,9 +92,7 @@ export const createBindingAtom = <T = unknown>(
             ? get(screenStorageAtom)
             : undefined,
         ),
-        user: rawJsExpression.includes("ensemble.user")
-          ? get(userAtom)
-          : undefined,
+        user: get(userAtom),
```
evshi, line 95: Can you revert this change? We don't want to create unnecessary depenencies.

packages/framework/src/evaluate/binding.ts
```diff
@@ -8,11 +8,11 @@ import {
   createStorageApi,
   screenStorageAtom,
 } from "../hooks/useEnsembleStorage";
+import { userAtom } from "../hooks/useEnsembleUser";
```
evshi, line 11: Let's keep the userAtom inside the state module. Hook can be its own file.

packages/framework/src/evaluate/binding.ts
sagardspeed2: so should I remove the get method from the useEmsembleUser hook ?

evshi: No, why would you do that? I'm just asking to revert these three lines here.

sagardspeed2: oh, I changed this while I changed user atom from state to hook. this is for the test only and it was need to be reverted back.

evshi (approved):

PR #399, opened 2024-02-08
fix: added blockage of unnecessary event passthrough on datarow

### LINKED ISSUE #396: DataRow's onTap action is fired when it's children's onTap should be
**Describe the bug**
In Datagrid widget, I have an onTap on DataRow and a Collapsible, among other widgets as its children. When I click on Collapsible widget, so both actions gets fired i.e., Collapsible's and DataRow's.

**To Reproduce**
Use this example yaml to reproduce the issue:

```yaml
- DataGrid:
      DataColumns:
        - label: Number
        - label: Name
      item-template:
        data: [1, 2]
        name: dummyData
        template:
          DataRow:
            onTap:
              navigateScreen:
                name: forms
            children:
              - Text:
                  text: ${dummyData}
              - Collapsible:
                  expandIconPosition: "end"
                  items:
                    - key: "2"
                      label: Here is my label 2
                      children:
                        Text:
                          text: Here is my value 2
                  onCollapse:
                    executeCode: |
                      console.log('HelloC', data)
```

**Expected behavior**

**Screenshots**

**Additional context**

sagardspeed2:
## Describe your changes
Added blockage of unnecessary event passthrough on datarow on tap

Closes #396 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Collapsible.tsx
+++ b/packages/runtime/src/widgets/Collapsible.tsx
@@ -188,14 +188,16 @@ export const Collapsible: React.FC<CollapsibleProps> = (props) => {
         },
       }}
     >
-      <Collapse
-        accordion={values?.isAccordion}
-        activeKey={activeValue}
-        expandIcon={expandIcon}
-        expandIconPosition={props.expandIconPosition}
-        items={collapsibleItems}
-        onChange={handleCollapsibleChange}
-      />
+      <div onClick={(e) => e.stopPropagation()}>
+        <Collapse
+          accordion={values?.isAccordion}
+          activeKey={activeValue}
+          expandIcon={expandIcon}
+          expandIconPosition={props.expandIconPosition}
+          items={collapsibleItems}
+          onChange={handleCollapsibleChange}
+        />
+      </div>
     </ConfigProvider>
   );
 };
--- a/packages/runtime/src/widgets/Collapsible.tsx
+++ b/packages/runtime/src/widgets/Collapsible.tsx
@@ -188,7 +188,12 @@ export const Collapsible: React.FC<CollapsibleProps> = (props) => {
         },
       }}
     >
-      <div onClick={(e) => e.stopPropagation()}>
+      <div
+        onClick={(e) => e.stopPropagation()}
+        onKeyDown={(e) => e.stopPropagation()}
+        role="button"
+        tabIndex={0}
+      >
         <Collapse
           accordion={values?.isAccordion}
           activeKey={activeValue}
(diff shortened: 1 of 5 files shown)


sagardspeed2:
@evshi I tried many other ways to stop event prapogation, also tried by the adding parent div wrapper around the collapsible. but none of them works. except this one.

Need your idea or suggestion on this one

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -198,13 +198,26 @@ export const DataGrid: React.FC<GridProps> = (props) => {
...
+  };
+
   return (
     <div ref={containerRef}>
       <Table
         dataSource={namedData}
         key={resolvedWidgetId}
         onRow={(record, recordIndex) => {
-          return { onClick: () => onTapActionCallback(record, recordIndex) };
+          return { onClick: (e) => handleOnRowClick(e, record, recordIndex) };
```
evshi, line 220: @sagardspeed2 I think you've misunderstood the issue. The issue is the event propagation from the Collapsible container. We need to wrap the Collapsible widget with a component and an event handler that stops event propagation.

evshi (changes requested):

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
sagardspeed2: @evshi I did not understand the task in wrong way. I tried your suggested way in which we can wrap the collapsible inside an element and using that element, we need to stop the event propagation. But that thing is not working. When we add event stop propagation inside the wrapper, it also block the click event on the collapsible widget. so on the click on collapsible, it does not trigger the collapsible callback.

evshi: @sagardspeed2 please double check what you're describing. It doesn't make sense that a container element would interfere with the event handling of its child element. I pushed a commit that resolves the issue with the approach I'm describing. Let me know how it differs from what you were seeing.

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -188,14 +188,21 @@ export const Collapsible: React.FC<CollapsibleProps> = (props) => {
...
-        expandIcon={expandIcon}
-        expandIconPosition={props.expandIconPosition}
-        items={collapsibleItems}
-        onChange={handleCollapsibleChange}
-      />
+      <div
+        onClick={(e) => e.stopPropagation()}
+        onKeyDown={(e) => e.stopPropagation()}
+        role="button"
+        tabIndex={0}
```
evshi, line 195: These aren't necessary. If you're getting ESLint errors, let's ignore them for now.

evshi (approved):

PR #403, opened 2024-02-11
toggle button styles update

sibteali786:
## Describe your changes
- updated a minor styling for toggle button

Closes #

- [ ] I have added tests


packages/runtime/src/widgets/ToggleButton.tsx
```diff
@@ -136,6 +136,7 @@ export const ToggleButton: React.FC<ToggleButtonProps> = ({
               backgroundColor: values?.styles?.selectedBackgroundColor,
               filter: "brightness(92%)",
             },
+            ...values?.styles,
```
evshi, line 139: I don't think the styles should be on the individual button elements here...the styles should be on the parent ButtonGroup element. I think each individual element should style itself to take up the remaining parent space in equal parts.

evshi (approved):

PR #417, opened 2024-02-14
add body attribute in showDialog action

### LINKED ISSUE #406: Support body attribute in showDialog like flutter
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

anserwaseem:

Closes #406

- [ ] I have added tests


packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -259,11 +259,11 @@ export const useShowDialog: EnsembleActionHook<ShowDialogAction> = (
     ensembleAction.callback();
   }, [ensembleAction]);
 
-  if (!action?.widget)
+  if (!action?.widget && !action?.body)
     throw new Error("ShowDialog Action requires a widget to be specified");
   const widget = useMemo(
-    () => unwrapWidget(cloneDeep(action.widget)),
-    [action.widget],
+    () => unwrapWidget(cloneDeep(action?.widget || action?.body || {})),
```
evshi, line 265: You can use ?? instead of || in typescript

packages/runtime/src/runtime/modal.tsx
```diff
@@ -219,6 +246,18 @@ export const ModalWrapper: React.FC<PropsWithChildren> = ({ children }) => {
     cursor: "pointer",
   };
 
+  const getFullHeight = (options?: ModalProps): string =>
```
evshi, line 249: What is the problem you are trying to solve for here? This seems very complex, and complex code is hard to maintain.

anserwaseem: there was hard-coded height being used in the code, I'm trying to get rid of that and set the height to 100 percent if it's given by the user

evshi: We need to decide which element is responsible for styling the sizing of the modal 
image

Seems like the parent element should receive the styling props and all inner elements should size itself to that. What do youthink?

anserwaseem: you are right but there is one caveat in this, we'd have to give body some height in one use case when `navigateModalScreen` is used and such screen is opened inside modal which has a sticky footer. For example, this one https://github.com/EnsembleUI/ensemble-react/blob/67f44d3dec369dede5c393f27ab1f9dee3ab9605/apps/kitchen-sink/src/ensemble/screens/actions.yaml#L113-L114

evshi: The body is always sized to parent element, and the footer is rendered over the top of the body. What would be the issue there?

packages/runtime/src/runtime/modal.tsx
```diff
@@ -89,7 +98,12 @@ export const ModalWrapper: React.FC<PropsWithChildren> = ({ children }) => {
       {
         visible: true,
         content: newContent,
-        options: newOptions,
+        options: evaluate(
+          screen as ScreenContextDefinition,
+          (isString(newOptions) ? newOptions : JSON.stringify(newOptions))
+            // eslint-disable-next-line prefer-named-capture-group
+            ?.replace(/['"]\$\{([^}]*)\}['"]/g, "$1"),
+        ),
```
evshi, line 106: Instead of using `evaluate` directly use `useEvaluate` like the `useNavigateScreen` hook. Requires less dependencies and is more performant.

packages/runtime/src/runtime/modal.tsx
```diff
@@ -83,13 +94,28 @@ export const ModalWrapper: React.FC<PropsWithChildren> = ({ children }) => {
       ]);
     }
 
+    let stringOptions = "";
+    if (isString(newOptions)) {
+      stringOptions = newOptions;
+    } else {
+      stringOptions = JSON.stringify({ ...newOptions, title: "" });
+    }
+    const evaluatedOptions = evaluate<ModalProps>(
```
evshi, line 103: Please avoid using evaluate directly and use `useEvaluate`.

evshi (changes requested):
Please address the usage of `evaluate` and the full height special handling. @sibteali786  please add more detail on what additional issue you are facing with the parent style name.

@anserwaseem in the future I highly recommend breaking this up into separate PRs since this has grown beyond the scope of the original ticket.

sibteali786:
Hey @evshi , I made a separate video to explain an issue I'm facing with a model not positioning correctly. Can you please watch it and give me your suggestions on how to solve it? https://www.loom.com/share/38f98e2fb73f4a03a1f0dfd9b56f3638?src=composer 
@anserwaseem let me know if i have correctly mentioned the problem.

packages/runtime/src/runtime/modal.tsx
anserwaseem: so, I'm applying full height on `.ant-modal-body` class for 2 occasions:
1. when we have full screen (i.e. `hideFullScreenIcon` is passed in as false and user taps on its icon to fully open the modal)
2. when user passes in full height using `styles` or `options` prop
For calculation, I'm just subtracting title bar height and given paddings to make the modal body of exact height that it should be.

packages/runtime/src/widgets/Icon.tsx
```diff
@@ -27,11 +27,19 @@ export const Icon: React.FC<IconProps> = ({ onTap, onHover, ...props }) => {
   if (!IconComponent) {
     return null;
   }
+  const handleMouseEnter = (
```
evshi, line 30: Remove these changes?

packages/runtime/src/runtime/modal.tsx
evshi: You should be able to achieve the same effect with minimal calculations. Consider the following structure:

```
parent div // responsible for height/width. creates a stacking context
  title/header div // responsible for containing header, can be fixed to top of parent stacking context OR inlined with content
  content div // contains body widget, sizes dynamically based on remaining space
  footer div // contains footer, fixed to bottom of parent
```

Given this structure, you should only need to calculate the offset of the content top margin based on header height.

evshi:
> Hey @evshi , I made a separate video to explain an issue I'm facing with a model not positioning correctly. Can you please watch it and give me your suggestions on how to solve it? https://www.loom.com/share/38f98e2fb73f4a03a1f0dfd9b56f3638?src=composer @anserwaseem let me know if i have correctly mentioned the problem.

This issue is a sign you're not taking the right approach and fighting the tools. I'll set up a meeting between the three of us to sort this out.

packages/runtime/src/runtime/modal.tsx
```diff
@@ -55,6 +65,16 @@ export const ModalWrapper: React.FC<PropsWithChildren> = ({ children }) => {
   const [modalDimensions, setModalDimensions] = useState<ModalDimensions[]>([]);
   const [isFullScreen, setIsFullScreen] = useState<boolean[]>([]);
   const contentRef = useRef<HTMLDivElement | null>(null);
+  const titleRef = useRef<HTMLDivElement | null>(null);
```
evshi, line 68: This ref is unused?

packages/runtime/src/runtime/modal.tsx
```diff
@@ -81,27 +126,16 @@ export const ModalWrapper: React.FC<PropsWithChildren> = ({ children }) => {
...
-    ]);
-
-    setIsFullScreen((oldIsFullScreen) => [...oldIsFullScreen, false]);
+    setContext(modalContext);
+    setNewContent(content);
+    setNewOptions(
+      omit(options, ["onClose", !isString(options.title) ? "title" : ""]),
+    );
+    setIsDialog(isADialog);
+    setIsDialogSet(false);
```
evshi, line 138: Does it make sense to just combine these state variables?

packages/runtime/src/shared/styles.ts
```diff
@@ -104,6 +104,7 @@ export const getIcon = (name: string): SvgIconComponent | undefined => {
 export const getComponentStyles = (
   name: string,
   styles?: React.CSSProperties,
+  addImportant = true,
```
evshi, line 107: Can we remove this?

packages/runtime/src/runtime/modal.tsx
```diff
@@ -188,29 +232,56 @@ export const ModalWrapper: React.FC<PropsWithChildren> = ({ children }) => {
...
-    .ensemble-modal-${index} .ant-modal-body {
-      height: calc(100vh - 110px);
+    .ensemble-modal-${index}, .ensemble-modal-${index} .ant-modal-content {
+      height: 100vh !important;
+      width: 100vw !important;
+      margin: 0 !important;
+      top: 0 !important;
+      left: 0 !important;
+      bottom: 0 !important;
+      right: 0 !important;
```
evshi, line 284: Why do you need these styles?

packages/runtime/src/runtime/modal.tsx
```diff
@@ -164,6 +207,7 @@ export const ModalWrapper: React.FC<PropsWithChildren> = ({ children }) => {
             ? getPositionSelector(
                 modalDimensions[index].width,
                 modalDimensions[index].height,
+                options.height?.includes("100"),
```
evshi, line 210: I don't think you even need the inline style selector you can pass styles directly to the modal to position. https://codesandbox.io/p/sandbox/to-customize-the-position-of-modal-antd-5-14-2-23xtnp?file=%2Fdemo.tsx%3A15%2C28

evshi (commented):
Can we please clean up the CSS and remove usages of important?

packages/runtime/src/runtime/modal.tsx
anserwaseem: I need these here when fullscreen is tapped in order to overwrite user given styles like height, top etc.

evshi: We can streamline this right? Since you're already calculating top/left/right/bottom and applying to styles prop you don't need separate CSS rules?

evshi (approved):

evshi:
@anserwaseem @sibteali786 to fix the onHover issue we'll need to position the modal wrapper relative to the location. So instead of the wrapper taking the entire screen, the wrapper will begin at the cursor position and the modal will position itself to the top left (this is going to also require extra calculations in responsiveness later on).

@sibteali786 let's keep the mouseenter and click out to close the modal for now and I created this ticket to resolve the onmouseleave. https://github.com/EnsembleUI/ensemble-react/issues/448

PR #423, opened 2024-02-15
flex widget implementation

### LINKED ISSUE #414: Add support for Flex

**Additional context**

sagardspeed2:
## Describe your changes
Added support for flex widget

Closes #414 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Flex.tsx
+++ b/packages/runtime/src/widgets/Flex.tsx
@@ -2,11 +2,13 @@ import {
   useRegisterBindings,
   type EnsembleWidget,
 } from "@ensembleui/react-framework";
+import { omit } from "lodash-es";
 import { WidgetRegistry } from "../registry";
-import {
-  type EnsembleWidgetStyles,
-  type EnsembleWidgetProps,
-  type HasItemTemplate,
+import type {
+  EnsembleWidgetStyles,
+  EnsembleWidgetProps,
+  HasItemTemplate,
+  FlexboxStyles,
 } from "../shared/types";
 import { Column } from "./Column";
 import { Row } from "./Row";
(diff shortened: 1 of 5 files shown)


packages/runtime/src/widgets/Flex.tsx
```diff
@@ -0,0 +1,91 @@
...
+  overflow?: string;
+}
+
+export type FlexProps = {
+  children?: EnsembleWidget[];
+} & HasItemTemplate &
+  EnsembleWidgetProps<FlexStyles>;
+
+export const FlexWidget: React.FC<FlexProps> = (props) => {
+  const { "item-template": itemTemplate, children, ...rest } = props;
```
evshi, line 32: Why not just render a Row/Column based on the direction property?

sagardspeed2: so you are suggesting instead of creating new widget, we can introduce the direction property on the row and column layout

evshi: No I mean keep this widget, just return the different component based on direction.

sagardspeed2: Something like, is direction is column then return the column widget, otherwise return the row widget ?

sagardspeed2:
@evshi so as per your commented, I updated this PR. now it will render column or row based on the given flex direction attributes.

but as per my suggestions, we need to stick with the previous implementation. because we are already providing the column and row widgets. so if customer want to use either column or row, then can directly use that without use of Flex widget.

share your thoughts on this.

packages/runtime/src/widgets/Flex.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const { values } = useRegisterBindings({ ...rest });
+
+  return (
+    <>
+      {values?.styles?.direction === "vertical" ? (
+        <Column
+          item-template={itemTemplate}
+          styles={{ gap: values.styles.gap || 4 }}
```
evshi, line 33: You need to pass the rest of the props

packages/runtime/src/widgets/Flex.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export type FlexProps = {
+  children?: EnsembleWidget[];
+} & HasItemTemplate &
+  EnsembleWidgetProps<FlexStyles>;
+
+export const FlexWidget: React.FC<FlexProps> = (props) => {
+  const { "item-template": itemTemplate, children, ...rest } = props;
+  const { values } = useRegisterBindings({ ...rest });
```
evshi, line 26: You don't need to call this since this is all handled by the underlying implementation

packages/runtime/src/widgets/Flex.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export type FlexProps = {
+  children?: EnsembleWidget[];
+} & HasItemTemplate &
+  EnsembleWidgetProps<FlexStyles & FlexboxStyles>;
+
+export const FlexWidget: React.FC<FlexProps> = (props) => {
+  const { values } = useRegisterBindings({
+    direction: props.styles?.direction,
+  });
```
evshi, line 29: You need to consider what happens if you call useRegisterBindings here when the Column and Row widgets also register their own bindings. Either remove this and just evaluate the direction OR you need to destructure the important bindings, like id.

evshi (approved):

PR #456, opened 2024-03-06
added support for expressions in screen url for navigate url

sibteali786:
## Describe your changes
- added support for url to be evaluated if an expression 

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/hooks/useNavigateUrl.ts
+++ b/packages/runtime/src/runtime/hooks/useNavigateUrl.ts
@@ -1,3 +1,4 @@
+import { evaluate } from "./../../../../framework/src/evaluate/evaluate";
 import { useMemo } from "react";
 import { useNavigate } from "react-router-dom";
 import {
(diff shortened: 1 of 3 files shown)


packages/runtime/src/runtime/hooks/useNavigateUrl.ts
```diff
@@ -42,7 +44,17 @@ export const useNavigateUrl: EnsembleActionHook<NavigateUrlAction> = (
         set(inputs, path, result);
       });
     }
-
+    if (isString(screenUrl) && isExpression(screenUrl) && screenContext) {
```
evshi, line 47: This should use `useEvaluate`

evshi (commented):
This can be cleaned up now so inputs are also passed through useEvaluate.

packages/runtime/src/runtime/hooks/useNavigateUrl.ts
```diff
@@ -2,13 +2,15 @@ import { useMemo } from "react";
...
-  evaluate,
   findExpressions,
   useCustomScope,
   useEnsembleStorage,
   useScreenContext,
+  isExpression,
+  useEvaluate,
 } from "@ensembleui/react-framework";
 import { cloneDeep, isString, set } from "lodash-es";
+import { evaluate } from "@ensembleui/react-framework/src/evaluate/evaluate";
```
evshi, line 13: Remove this please. Only use the top level export.

PR #462, opened 2024-03-11
Stepper Feature

sibteali786:
## Describe your changes
- Trying to fix activeStepIndex to support expressions 

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Stepper/Stepper.tsx
+++ b/packages/runtime/src/widgets/Stepper/Stepper.tsx
@@ -80,7 +80,6 @@ const Stepper: React.FC<StepperProps> = (props) => {
       handleBack,
     },
   );
-  console.log("Stepper", values);
   useEffect(() => {
     if (
       isString(props?.activeStepIndex) &&
--- a/packages/runtime/src/widgets/Stepper/Stepper.tsx
+++ b/packages/runtime/src/widgets/Stepper/Stepper.tsx
@@ -53,34 +53,43 @@ interface CustomConnectorProps {
 }
 
 const Stepper: React.FC<StepperProps> = (props) => {
-  const [activeStep, setActiveStep] = useState(props?.activeStepIndex);
+  const [activeStep, setActiveStep] = useState(props.activeStepIndex ?? 0);
 
   const itemTemplate = props["item-template"];
   const { namedData } = useTemplateData({ ...itemTemplate });
   const stepsLength = namedData.length;
   const stepTypes = itemTemplate?.template;
   const handleNext = useCallback(() => {
-    if (activeStep && activeStep < namedData.length - 1) {
-      setActiveStep(activeStep + 1);
+    if (activeStep < namedData.length - 1) {
+      setActiveStep((prevStep) => prevStep + 1);
     }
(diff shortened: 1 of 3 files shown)


sibteali786:
@evshi need some help here, first when I update step using Next it does not work, and remains on old value ? Second is it a right approach what i am doing here to make a separate hook for this ?

sibteali786:
The example is at `/widgets` screen at bottom

evshi (commented):
I think what you're looking for is just a `useEffect` hook that syncs the evaluated attribute to the component local state

packages/runtime/src/widgets/Stepper/Stepper.tsx
```diff
@@ -51,35 +54,39 @@ interface CustomConnectorProps {
 }
 
 const Stepper: React.FC<StepperProps> = (props) => {
-  const [activeStep, setActiveStep] = useState(props.activeStepIndex ?? 0);
+  // Explicit update to evaluate activeStepIndex
+  const resVals = useEvaluate({
+    activeStepIndex: props.activeStepIndex,
+  });
```
evshi, line 60: Remove this. Just move the useEffect hook after useRegisterBindings.

packages/runtime/src/widgets/Stepper/Stepper.tsx
```diff
@@ -32,7 +35,7 @@ export interface StepProps {
 
 export type StepperProps = {
   steps: StepProps[];
-  activeStepIndex?: number;
```
evshi, line 35: This can still be null?

packages/runtime/src/widgets/Stepper/Stepper.tsx
```diff
@@ -51,35 +53,35 @@ interface CustomConnectorProps {
...
-      handleNext,
-      handleBack,
-    },
-  );
+  const { values, rootRef } = useRegisterBindings({ ...props }, props.id, {
+    handleNext,
+    handleBack,
+  });
+  useEffect(() => {
+    if (isString(activeStep) && isExpression(activeStep)) {
```
evshi, line 80: You can check the props value so you don't need to overload your state variable. Your state variable should always just be an integer

evshi (approved):

packages/runtime/src/widgets/Stepper/Stepper.tsx
```diff
@@ -16,14 +16,16 @@ import {
   useRegisterBindings,
   useTemplateData,
   unwrapWidget,
+  isExpression,
 } from "@ensembleui/react-framework";
 import type { EnsembleWidget, Expression } from "@ensembleui/react-framework";
 import Alert from "@mui/material/Alert";
+import { isString } from "antd/es/button";
```
evshi, line 23: wrong import use from lodash-es

evshi (changes requested):

sibteali786:
there is one more issue happening here @evshi i forgot to discuss it yesterday, the state of stepper is not being updated when we change a step, it does change step but not the value of activeStepIndex nor stiveStep . This way it always stays what it was initialized with ?

PR #466, opened 2024-03-13
fix: added execution support of onResponse and onError in API block d…

### LINKED ISSUE #465: onResponse of the apis under API block are not working
**Describe the bug**
`onResponse` of the apis under API block of the screen are not working.

**To Reproduce**
Checkout this yaml
```yaml
View:
  onLoad:
    executeCode: |
      ensemble.invokeAPI('getDummyProducts');
  body:
    Text:
      text: This is a test

API:
  getDummyProducts:
    method: GET
    uri: https://dummyjson.com/products
    onResponse:
      executeCode: |
        console.log('response');
```

**Expected behavior**
`onResponse` working

sagardspeed2:
Added support for execution of onResponse and onError block of API block definition in sequential with invokeAPI's onResponse and onError

Closes #465 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
+++ b/packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
@@ -205,17 +205,19 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
   const { apis, setData } = useScreenData();
   const [isComplete, setIsComplete] = useState<boolean>();
   const [isLoading, setIsLoading] = useState<boolean>();
-
   const [context, setContext] = useState<unknown>();
-
   const evaluatedInputs = useEvaluate(action?.inputs, { context });
-  const onResponseAction = useEnsembleAction(action?.onResponse);
-  const onErrorAction = useEnsembleAction(action?.onError);
 
   const api = useMemo(
     () => apis?.find((model) => model.name === action?.name),
     [action?.name, apis],
   );
+
+  const onInvokeAPIResponseAction = useEnsembleAction(action?.onResponse);
+  const onInvokeAPIErrorAction = useEnsembleAction(action?.onError);
+
+  const onAPIResponseAction = useEnsembleAction(api?.onResponse);
+  const onAPIErrorAction = useEnsembleAction(api?.onError);
 
   const invokeApi = useMemo(() => {
     if (!api) {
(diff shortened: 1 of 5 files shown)


packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -205,18 +205,20 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
...
   const api = useMemo(
     () => apis?.find((model) => model.name === action?.name),
     [action?.name, apis],
   );
 
+  const onInvokeAPIResponseAction = useEnsembleAction(action?.onResponse);
+  const onInvokeAPIErrorAction = useEnsembleAction(action?.onError);
+
+  const onAPIResponseAction = useEnsembleAction(api?.onResponse);
+  const onAPIErrorAction = useEnsembleAction(api?.onError);
```
evshi, line 220: Don't call hook multiple times per callback. Just get the correct callback model once.

evshi (changes requested):

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
evshi: Discussed offline, both are needed if there are multiple onResponse for onLoad and on the API definition. However we still need to handle this for the code `invokeAPI` case.

sagardspeed2: This applies to invokeAPI too. The API is only called when you use invokeAPI. Even during onLoad, invokeAPI is used, triggering the API in that case.

evshi: I meant when you call ensemble.invokeAPI from code? It works there too?

sagardspeed2: No, in `ensemble.invokeAPI` we are not passing any onResponse or onError callback. so it is not intended to work there

evshi: Right, but when you call `ensemble.invokeAPI`, should the `onResponse` of the API model fire?

sagardspeed2: No

anserwaseem: > Right, but when you call `ensemble.invokeAPI`, should the `onResponse` of the API model fire?

It should

evshi: I'll add this in a separate PR

evshi (approved):

PR #474, opened 2024-03-18
fix: only render dropdown after options render

### LINKED ISSUE #470: Dropdown value is not initialized correctly
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

sagardspeed2:
## Describe your changes
to fix delay in initial loaded value with the right option, only allow dropdown render once options are loaded

Closes #470 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Form/Dropdown.tsx
+++ b/packages/runtime/src/widgets/Form/Dropdown.tsx
@@ -152,6 +152,10 @@ const Dropdown: React.FC<DropdownProps> = (props) => {
 
   const { backgroundColor: _, ...formItemStyles } = values?.styles ?? {};
 
+  if (isEmpty(options)) {
+    return null;
+  }
+
   return (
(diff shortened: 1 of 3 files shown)


packages/runtime/src/widgets/Form/Dropdown.tsx
```diff
@@ -152,6 +152,10 @@ const Dropdown: React.FC<DropdownProps> = (props) => {
 
   const { backgroundColor: _, ...formItemStyles } = values?.styles ?? {};
 
+  if (isEmpty(options)) {
```
evshi, line 155: Maybe check for `null` not just emptiness. Sometimes an empty list is a valid use case.

evshi (commented):
Please add a kitchen sink example

evshi (approved):

PR #486, opened 2024-03-22
fix: added id attribute on formItem widget

### LINKED ISSUE #485: TextInput.setValue doesn't work
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

sagardspeed2:
## Describe your changes
Added support of id attribute on formItem widget to set form elements value through setValue

Closes #485 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Form/FormItem.tsx
+++ b/packages/runtime/src/widgets/Form/FormItem.tsx
@@ -1,9 +1,9 @@
 import type { FormItemProps } from "antd";
 import { Form as AntForm } from "antd";
-import type { FormInputProps } from "./types";
 import { isString } from "lodash-es";
-import { EnsembleRuntime } from "../../runtime";
 import { unwrapWidget } from "@ensembleui/react-framework";
+import { EnsembleRuntime } from "../../runtime";
+import type { FormInputProps } from "./types";
 
 export type EnsembleFormItemProps<T> = FormItemProps & {
   values?: FormInputProps<T>;
@@ -17,9 +17,11 @@ export const EnsembleFormItem: React.FC<EnsembleFormItemProps<unknown>> = (
 ) => {
   const { values, ...rest } = props;
   const { backgroundColor: _, ...formItemStyles } = values?.styles ?? {};
+
   return (
     <AntForm.Item
       className={values?.styles?.names}
+      id={values?.id ?? values?.label}
       initialValue={values?.value}
       label={
         values?.label ? (
@@ -36,7 +38,6 @@ export const EnsembleFormItem: React.FC<EnsembleFormItemProps<unknown>> = (
           </label>
         ) : null
       }
-      name={values?.id ?? values?.label}
       rules={[{ required: Boolean(values?.required) }]}
       style={{
         margin: "0px",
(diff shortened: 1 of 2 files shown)


packages/runtime/src/widgets/Form/FormItem.tsx
```diff
@@ -17,9 +17,11 @@ export const EnsembleFormItem: React.FC<EnsembleFormItemProps<unknown>> = (
 ) => {
   const { values, ...rest } = props;
   const { backgroundColor: _, ...formItemStyles } = values?.styles ?? {};
+
   return (
     <AntForm.Item
       className={values?.styles?.names}
+      id={values?.id ?? values?.label}
```
evshi, line 24: I don't think this is the issue...can you create a kitchen sink example?

sagardspeed2: I added a example in kitchen-sink for this, where you can check this

evshi: Hmmm can you explain to me what the issue was and how this resolves this?

sagardspeed2: @evshi https://ant.design/components/input
according to document, when we wrap any form elements with the form item, then form item take all the control of the form element, and the value, and id prop of the form elements are controlled by the form item only.

anserwaseem: @sagardspeed2 but your change (removing `name` control) caused Form not to be submitted anymore.

packages/runtime/src/widgets/Form/TextInput.tsx
```diff
@@ -21,6 +21,11 @@ export const TextInput: React.FC<TextInputProps> = (props) => {
   const { values } = useRegisterBindings({ ...props, value }, props.id, {
     setValue,
   });
+  const form = Form.useFormInstance();
+
+  useEffect(() => {
+    form.setFieldsValue({ [values?.id ?? values?.label ?? ""]: value });
+  }, [value]);
```
evshi, line 28: We might not always be in a Form instance...this should still work? Why not just call setValue and then onChange?

packages/runtime/src/widgets/Form/FormItem.tsx
evshi: @sagardspeed2 please explain why the issue occurs not just point at documentation. Only if you understand the root cause can you create a proper fix.

packages/runtime/src/widgets/Form/TextInput.tsx
```diff
@@ -21,11 +21,15 @@ export const TextInput: React.FC<TextInputProps> = (props) => {
...
 
   useEffect(() => {
-    form.setFieldsValue({ [values?.id ?? values?.label ?? ""]: value });
-  }, [value]);
+    if (formInstance) {
+      formInstance.setFieldsValue({
+        [values?.id ?? values?.label ?? ""]: value,
+      });
+    }
+  }, [value, formInstance]);
```
evshi, line 32: Calling onChange didn't work?

packages/runtime/src/widgets/Form/FormItem.tsx
evshi: I removed this since it's not necessary for the solution to work fyi @sagardspeed2

evshi (approved):

PR #488, opened 2024-03-22
fix: added support of expressions in style and class names

### LINKED ISSUE #483: Style names don't support expressions
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

sagardspeed2:
## Describe your changes
Added support of expressions in style and class names

Closes #483 

- [ ] I have added tests


packages/framework/src/hooks/useStyleNames.ts
```diff
@@ -1,14 +1,60 @@
...
+): CSSProperties => {
+  let names: string[] = isString(styleNames)
+    ? styleNames.split(" ")
+    : styleNames;
+
+  names = names.map((styleName) => {
+    const isClass = styleName.startsWith(".");
+    const styleKey = isClass ? styleName.substring(1) : styleName;
+    const styleVal = isExpression(styleKey)
+      ? evaluate<string>(defaultScreenContext, styleKey, {
```
evshi, line 28: Don't use `evaluate` directly. Did you see the warning? Use `useEvaluate` and then update the dependency array of the second evaluation to include styles like we discussed.

evshi (changes requested):

anserwaseem (approved):

packages/framework/src/hooks/useStyleNames.ts
```diff
@@ -1,28 +1,56 @@
...
+  // evaluate classnames
+  const classStylesArray: string[] | undefined = isString(classStyles)
+    ? classStyles.split(" ")
+    : classStyles;
+  const styleClasses = useEvaluate({ classStylesArray });
+
+  const styleNames = [
+    ...(nameStyles.nameStylesArray
+      ? nameStyles.nameStylesArray.map((nameStyle) => `${nameStyle}`)
+      : []),
```
evshi, line 47: You don't need both the spread operator and map. Both will create new arrays.

packages/framework/src/hooks/useStyleNames.ts
```diff
@@ -1,28 +1,56 @@
...
+  const nameStylesArray: string[] | undefined = isString(namedStyles)
+    ? namedStyles.split(" ")
+    : namedStyles;
+  const nameStyles = useEvaluate({ nameStylesArray });
+
+  // evaluate classnames
+  const classStylesArray: string[] | undefined = isString(classStyles)
+    ? classStyles.split(" ")
+    : classStyles;
+  const styleClasses = useEvaluate({ classStylesArray });
```
evshi, line 42: You can evaluate both in a single call - any reason to split them up here?

packages/framework/src/parser.ts
```diff
@@ -348,9 +348,15 @@ const unwrapTheme = (theme?: string): EnsembleThemeModel | undefined => {
...
-  );
+  resolvedProps.forEach(([path, value]) => {
+    const stylePaths = path.split(".");
+    if (path.startsWith(".")) {
+      stylePaths.shift();
+      stylePaths[0] = `.${stylePaths[0]}`;
+    }
+
+    set(workingTheme.Styles!, [...stylePaths], value);
+  });
```
evshi, line 359: I still don't think I understand what this is doing. Can you add an example here? Or even better, a unit test?

packages/framework/src/hooks/useStyleNames.ts
sagardspeed2: Yes, because on classStyle evaluation, we need to add prefix '.' that indicate that this one is class style

packages/framework/src/hooks/useStyleNames.ts
sagardspeed2: this is for merge the names styles and class styles in single array

packages/framework/src/parser.ts
sagardspeed2: this is very important for evaluate class styles with expressions
for example,

if theme file like,

```
{
    Styles: {
        .button-1: {
            backgroundColor: "${colors.primary['500']}"
        }
    }
}
```

and for evaluate this style correctly we need to convert them into
```
{
    Styles: {
        .button-1: {
            backgroundColor: "#ff0"
        }
    }
}
```

the logic here written is for set the value in object properties where key name starts with '.'

evshi: How is it a complex expression? What happens if the expression of the style is `${someValue ?? otherValue}`

evshi:
Per discussion standup, let's add some unit tests for the parser changes to clarify what its purpose is.

apps/kitchen-sink/src/App.tsx
```diff
@@ -1,7 +1,7 @@
 import type { ApplicationDTO } from "@ensembleui/react-framework";
 import { EnsembleApp } from "@ensembleui/react-runtime";
 // Screens
-import MenuYAML from "./ensemble/screens/menu.yaml";
+// import MenuYAML from "./ensemble/screens/menu.yaml";
```
evshi, line 4: Please clean this up to original state

packages/framework/src/hooks/useEvaluate.ts
```diff
@@ -21,7 +21,7 @@ export const useEvaluate = <T extends Record<string, unknown>>(
       findExpressions(values, [], expressionMap);
       return expressionMap;
     },
-    options?.refreshExpressions ? [values] : [],
+    options?.refreshExpressions ? [values] : [values?.styles],
```
evshi, line 24: Let's leave a comment here - `styles` should be an object that represents a set of style rules. I wonder if this is maybe something we can extend to be more generic and not be special to styles...

packages/framework/src/hooks/useRegisterBindings.ts
```diff
@@ -40,12 +40,20 @@ export const useRegisterBindings = <T extends { [key: string]: unknown }>(
...
-  const styles = merge({}, styleProperties, values.styles);
-  if (!isEmpty(styles)) {
-    merge(values, { styles });
-  }
+  const styles = useMemo(() => {
+    return merge(
+      {},
+      styleProperties,
+      omit(values.styles ?? {}, ["className", "names"]),
+    );
```
evshi, line 48: Instead of creating another hook here - let's combine this logic inside `useStyleNames` and rename the hook to `useStyles`

packages/framework/src/hooks/useStyleNames.ts
```diff
@@ -1,28 +1,66 @@
...
+
+  const appliedStyles = {};
+  overlappingNames.forEach((name) => {
+    merge(appliedStyles, get(theme.Styles, name));
+  });
+
+  return appliedStyles;
+};
 
 export const useStyleNames = (
```
evshi, line 26: Rename to `useStyleSet` or something that better describes what this does.

packages/framework/src/hooks/useStyleNames.ts
```diff
@@ -1,28 +1,66 @@
...
+      namedStyles.split(/([a-zA-Z0-9_-]+)\s+(\${.+?})\s+(\w+)/).filter((x) => x)
+    : namedStyles;
 
-  if (styleNames && themeContext) {
-    return resolveStyleNames(styleNames, themeContext);
-  }
+  // evaluate classnames
+  const classStylesArray: string[] | undefined = isString(classStyles)
+    ? // eslint-disable-next-line prefer-named-capture-group
+      classStyles.split(/([a-zA-Z0-9_-]+)\s+(\${.+?})\s+(\w+)/)
```
evshi, line 41: Can we reuse the same regex everywhere, i.e. from fetcher.ts?

packages/framework/src/hooks/useStyleNames.ts
```diff
@@ -1,28 +1,66 @@
...
+    nameStylesArray: namedStylesArrayEval,
+  } = useEvaluate({
+    classStylesArray,
+    nameStylesArray,
+  });
+
+  const styleNames = [
+    ...(namedStylesArrayEval
+      ?.filter((className) => className)
+      .map((nameStyle) => `${nameStyle}`) || []),
```
evshi, line 55: You can just use lodash.compact to remove nulls/undefined. It will also create a new array.

evshi (commented):
Glad we got this to the right solution. Let's clean up the code

packages/framework/src/hooks/useStyleNames.ts
sagardspeed2: No, because this regex is for split string, so we can convert string into array of classname or stylename, and that classname or stylename, may be expression or not.

and the regex added in fetcher is for checking that given string is expression or not.

packages/framework/src/hooks/useEvaluate.ts
evshi: @sagardspeed2 please leave a comment here

packages/framework/src/hooks/useStyles.ts
```diff
@@ -0,0 +1,81 @@
...
+    nameStylesArray: namedStylesArrayEval,
+  } = useEvaluate({
+    classStylesArray,
+    nameStylesArray,
+  });
+
+  const styleNames = compact([
+    ...namedStylesArrayEval.map((nameStyle) => `${nameStyle}`),
+    ...classStylesArrayEval.map((className) => `.${className}`),
+  ]).join(" ");
```
evshi, line 66: This can be moved inside the next useMemo and there's no point joining and resplitting.

packages/framework/src/hooks/useStyles.ts
```diff
@@ -0,0 +1,81 @@
...
+  const {
+    classStylesArray: classStylesArrayEval,
+    nameStylesArray: namedStylesArrayEval,
+  } = useEvaluate({
+    classStylesArray,
+    nameStylesArray,
+  });
+
+  const styleNames = compact([
+    ...namedStylesArrayEval.map((nameStyle) => `${nameStyle}`),
```
evshi, line 64: This map is not required.

packages/framework/src/hooks/useStyles.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  const { classStylesArray, nameStylesArray } = useMemo(() => {
+    const stringSplitter = (styles: string): string[] =>
+      // eslint-disable-next-line prefer-named-capture-group
+      styles.split(/([a-zA-Z0-9_-]+)\s+(\${.+?})\s+(\w+)/).filter(Boolean);
+    return {
+      classStylesArray: isString(classStyles)
+        ? stringSplitter(classStyles)
+        : [],
+      nameStylesArray: isString(namedStyles) ? stringSplitter(namedStyles) : [],
```
evshi, line 51: Does this still work if there are mixed static and expressions, for example

```yaml
styles:
  className: "myStyle ${ensemble.storage.get('myOtherStyle')}"
```

I think a better way to do this is to just treat the entire thing as a string interpolation.

evshi (changes requested):
@sagardspeed2 there is still lots to clean up.

packages/framework/src/hooks/useStyles.ts
sagardspeed2: Yes, and we just can't treat this as entire string interpolation, as due to classname we need to add '.' as prefix on each classname

packages/framework/src/hooks/useStyles.ts
sagardspeed2: No that is wrong, we need to place this outside of useMemo and convert it into string and later resplitting. otherwise this will go into infinite render due to array gives new reference each time.

packages/framework/src/hooks/useStyles.ts
```diff
@@ -35,40 +34,28 @@ const resolveStyleNames = (
...
+      namedStyles: `\${\`${namedStyles || ""}\`}`,
+    });
 
   const styleNames = compact([
-    ...namedStylesArrayEval,
-    ...classStylesArrayEval.map((className) => `.${className}`),
+    ...(classStylesEval.trim()
+      ? classStylesEval.split(" ").map((className) => `.${className}`)
+      : []),
+    ...[namedStylesEval.trim()],
```
evshi, line 52: Don't need to split here?

sagardspeed2: We need to assign `.` prefix  to each classname, and this is the only reason to split here

evshi (approved):

PR #490, opened 2024-03-22
Handle nullish label in DataColumn

evshi:

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
+++ b/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
@@ -14,16 +14,9 @@ import {
   type EnsembleAction,
   unwrapWidget,
 } from "@ensembleui/react-framework";
-import {
-  useCallback,
-  type ReactElement,
-  useState,
-  useMemo,
-  useRef,
-  useEffect,
-  ReactEventHandler,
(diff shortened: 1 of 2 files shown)


packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -237,11 +230,15 @@ export const DataGrid: React.FC<GridProps> = (props) => {
               : {}),
           };
         }
+        let label = null;
+        if (isString(item.label)) {
+          label = item.label;
+        } else if (isObject(item.label)) {
+          EnsembleRuntime.render([unwrapWidget({ ...item.label })]);
```
anserwaseem, line 237: I think 
1. we need `cloneDeep` on item.label as spread operator won't deep copy nested objects
2. assign `render` functions's output to label var

anserwaseem (changes requested):

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
evshi: Good catch, ty.

anserwaseem (approved):

PR #494, opened 2024-03-26
Add force option to useRegisterBindings hook

anserwaseem:
## Describe your changes
Add force option in useRegisterBindings hook

### Problem
React component using useRegisterBindings hook was not being re-rendered properly when re-mounted with same values. 

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/hooks/useRegisterBindings.ts
+++ b/packages/framework/src/hooks/useRegisterBindings.ts
@@ -19,6 +19,9 @@ export const useRegisterBindings = <T extends { [key: string]: unknown }>(
   values: T,
   id?: string,
   methods?: InvokableMethods,
+  options?: {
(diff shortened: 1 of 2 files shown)


packages/framework/src/hooks/useRegisterBindings.ts
```diff
@@ -19,6 +19,9 @@ export const useRegisterBindings = <T extends { [key: string]: unknown }>(
   values: T,
   id?: string,
   methods?: InvokableMethods,
+  options?: {
+    forceRerender?: boolean;
+  },
```
evshi, line 24: Let's rename this to `forceState` and make it check explicitly for `true` value. You're going to have to be careful with using this in your component.

evshi (approved):

PR #500, opened 2024-04-02
Debounce state changes from TextInput

evshi:

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Form/TextInput.tsx
+++ b/packages/runtime/src/widgets/Form/TextInput.tsx
@@ -1,7 +1,9 @@
 import { Input, Form } from "antd";
 import type { Expression } from "@ensembleui/react-framework";
 import { useRegisterBindings } from "@ensembleui/react-framework";
-import { useEffect, useState } from "react";
+import type { ChangeEvent } from "react";
+import { useEffect, useMemo, useState } from "react";
+import { debounce } from "lodash-es";
 import type { EnsembleWidgetProps } from "../../shared/types";
 import { WidgetRegistry } from "../../registry";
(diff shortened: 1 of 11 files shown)


packages/runtime/src/widgets/Form/TextInput.tsx
```diff
@@ -39,12 +41,22 @@ export const TextInput: React.FC<TextInputProps> = (props) => {
...
 
+  const changeHandler = (
+    event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>,
+  ): void => setValue(event.target.value);
+
+  // We need to debounce the change handler otherwise state updates will create too many re-renders
+  const debouncedChangeHandler = useMemo(
+    () => debounce(changeHandler, 300),
+    [],
+  );
```
sagardspeed2, line 52: @evshi This is good to add debounce in this textInput to avoid frequently unnecessary renders of the component, but at same time, this is very bad experience on typing any values in textInput. If I want to type anything on input, I can't type at my regular speed, I need to press one key, then wait for some time and then I need to press other key, overall to work this, we need to reduce the typing speed. and for end user, this is not an ideal behaviour

sagardspeed2 (changes requested):

packages/runtime/src/widgets/Form/TextInput.tsx
evshi: Yeah this was meant to only debounce the widget state value, not the UI I need to adjust this.

evshi: Resolved this by moving the debounce to useRegisterBindings with an option to control.

PR #502, opened 2024-04-02
fix: receive updated values from form elements

sagardspeed2:
## Describe your changes
Receive updated values from form elements

Closes #

- [ ] I have added tests

### DIFF
--- a/apps/kitchen-sink/src/ensemble/screens/forms.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/forms.yaml
@@ -122,13 +122,107 @@ View:
                   id: result
                   text: Check here for the result!
 
-        - Button:
-            label: set text input value
+        - Text:
             styles:
-              margin: 5px
-            onTap:
-              executeCode: |
-                console.log('>>>>', radio)
+              names: heading-3
+            text: Forms with initial values
+        - Markdown:
+            text: More to come! In the meantime, checkout the Ensemble [documentation](https://docs.ensembleui.com/).
+        - Card:
+            styles:
+              maxWidth: unset
+              width: unset
+            children:
+              - Form:
+                  id: myForm2
+                  styles:
+                    gap: 8px
+                    labelPosition: top
+                  onSubmit:
+                    executeCode:
+                      body: |
+                        const formattedVals = JSON.stringify(vals, null, 2)
+                        console.log(formattedVals)
+                        result.setText(formattedVals)
+                      onComplete:
+                        showToast:
+                          message: Form submitted!
+                  children:
+                    - Radio:
+                        id: initial_radio
+                        label: Radio Group
+                        value: 2
+                        items:
+                          - label: Option 1
+                            value: 1
+                          - label: Option 2
+                            value: 2
+                          - label: Option 3
+                            value: 3
+                        onChange:
+                          executeCode: |
+                            console.log('radio changed', value)
+                    # Build your own checkbox group
+                    - Row:
+                        children:
+                          - Checkbox:
+                              id: initial_checkbox
+                              value: true
+                              trailingText: Check me
+                              onChange:
+                                executeCode: |-
+                                  console.log('isChecked', value);
+                          - Checkbox:
+                              id: initial_checkbox2
+                              trailingText: Check me 2
+                    - Dropdown:
+                        id: initail_dropdown
+                        label: Dropdown
+                        value: option2
+                        hintText: Choose One
+                        items:
+                          - label: Option 1
+                            value: option1
+                          - label: Option 2
+                            value: option2
+                    # If you omit id, the form value key will be the label
+                    - TextInput:
+                        id: initial_formTextInput
+                        value: ${ensemble.storage.get('inputVal')}
+                        label:
+                          Text:
+                            text: Text input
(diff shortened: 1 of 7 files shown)


packages/runtime/src/widgets/Form/Checkbox.tsx
```diff
@@ -53,6 +53,10 @@ export const CheckboxWidget: React.FC<CheckBoxProps> = (props) => {
 
   const formInstance = Form.useFormInstance();
 
+  useEffect(() => {
+    setChecked(values?.initialValue as boolean);
```
evshi, line 57: Instead of casting, you can use the `Boolean` constructor which will actually convert it to a boolean.

packages/runtime/src/widgets/Form/Date/Date.tsx
```diff
@@ -51,15 +55,23 @@ export const Date: React.FC<DateProps> = (props) => {
...
     } else {
       onChangeCallback(formattedDate);
     }
   };
 
+  useEffect(() => {
+    if (!value) {
+      setValue(
+        values?.initialValue || (dayjs(values?.value) as unknown as string),
+      );
```
evshi, line 71: Why would you not convert the initialDate also?

packages/runtime/src/widgets/Form/Date/Date.tsx
```diff
@@ -51,15 +55,23 @@ export const Date: React.FC<DateProps> = (props) => {
   );
 
   const onDateChange = (date: unknown): void => {
+    setValue(date as string);
```
evshi, line 58: Just change the argument to type date?

packages/runtime/src/widgets/Form/MultiSelect.tsx
```diff
@@ -47,21 +47,28 @@ const MultiSelect: React.FC<MultiSelectProps> = (props) => {
...
     },
   );
 
+  useEffect(() => {
+    if (!selectedValues) {
+      const initVal = isString(values?.initialValue)
+        ? [values?.initialValue]
+        : values?.initialValue;
+      setSelectedValues(initVal as string[]);
+    }
```
evshi, line 69: I think for this one `selectedValues` is the initial value.

sagardspeed2: No, actually I did not assign the initial value in local state while state initiate, we are updating state value after initial value from the props are being evaluated by the `useRegisterBindings` only

evshi: I meant that `initialValue` should be derived from `selectedValues` not the `value` prop.

sagardspeed2: But on the first render, `selectedValues` will be empty array and we will assign the initialValues,  and when we change the any value of `Multiselect`, this time we only need to update the selectedValues, no need to reassign.

apps/kitchen-sink/src/ensemble/screens/forms.yaml
```diff
@@ -122,13 +122,107 @@ View:
...
+                  id: myForm2
+                  styles:
+                    gap: 8px
+                    labelPosition: top
+                  onSubmit:
+                    executeCode:
+                      body: |
+                        const formattedVals = JSON.stringify(vals, null, 2)
+                        console.log(formattedVals)
+                        result.setText(formattedVals)
```
evshi, line 146: The result currently shows up in the other form.

packages/runtime/src/widgets/Form/MultiSelect.tsx
```diff
@@ -47,21 +47,28 @@ const MultiSelect: React.FC<MultiSelectProps> = (props) => {
...
       setSelectedValues,
       setOptions,
     },
   );
 
+  useEffect(() => {
+    if (!selectedValues) {
+      const initVal = isString(values?.initialValue)
+        ? [values?.initialValue]
+        : values?.initialValue;
```
evshi, line 67: If this doesn't evaluate to an array, maybe we skip setting the initial value.

packages/runtime/src/widgets/Form/Date/Date.tsx
```diff
@@ -50,16 +54,25 @@ export const Date: React.FC<DateProps> = (props) => {
...
       onChangeCallback(formattedDate);
     }
   };
 
+  useEffect(() => {
+    if (!value) {
+      setValue(
+        values?.initialValue ||
+          (dayjs(values?.initialValue) as unknown as string),
+      );
```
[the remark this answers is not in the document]
sagardspeed2: Yes

packages/runtime/src/widgets/Form/Date/Date.tsx
```diff
@@ -84,7 +94,7 @@ export const Date: React.FC<DateProps> = (props) => {
         onChange={onDateChange}
         placeholder={values?.hintText}
         style={{ width: "100%", ...values?.styles }}
-        value={dayjs(values?.value)}
+        value={String(dayjs(values?.value))}
```
evshi, line 97: Why is this change needed?

evshi (approved):

PR #510, opened 2024-04-05
fix: added support for onPageChange action in datagrid

### LINKED ISSUE #509: Support ensemble actions on data grid page selection
For data grids with pages, we need to support actions when users select pages so they can load more data if needed. This is similar to the use case of onScrollEnd

**Additional context**

sagardspeed2:
## Describe your changes
Added support for onPageChange action handler in datagrid

Closes #509 

- [ ] I have added tests

### DIFF
--- a/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/widgets.yaml
@@ -2,14 +2,17 @@ View:
   onLoad:
     executeCode:
       body: |-
-        ensemble.storage.set('dataColumn', Array(15).fill().map((x,i) => ({key1: 'value', key2: i})))
         ensemble.storage.set('activeIndex',3);
       onComplete:
         invokeApi:
-          name: getDummyNumbers
+          name: getDummyProductsByPaginate
+          inputs:
+            skip: 0
           onResponse:
             executeCode: |
-              ensemble.storage.set('dummyNumbers', response.body)
+              const res = response.body.products;
+              ensemble.storage.set("skip", 10)
+              ensemble.storage.set("dummyProducts", res);
 
   header:
     title:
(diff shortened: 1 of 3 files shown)


packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -263,6 +269,22 @@ export const DataGrid: React.FC<GridProps> = (props) => {
...
 
+  const onPageChangeAction = useEnsembleAction(onPageChange);
+  // page change action
+  const onPageChangeActionCallback = useCallback(
+    (page: number) => {
+      if (onPageChangeAction) {
+        onPageChangeAction.callback({ page });
+      }
+    },
+    [onPageChangeAction],
```
evshi, line 280: We can also pass the page size

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -84,6 +84,7 @@ export type GridProps = {
   hidePagination?: boolean;
   scroll?: DataGridScrollable;
   onScrollEnd?: EnsembleAction;
+  onPageChange?: EnsembleAction;
```
evshi, line 87: Let's add an attribute for pageSize and if they pass 0 or less it's equivalent to hiding pagination

apps/kitchen-sink/src/ensemble/screens/widgets.yaml
```diff
@@ -2,7 +2,7 @@ View:
   onLoad:
     executeCode:
       body: |-
-        ensemble.storage.set('dataColumn', [{key1: "value", key2: 2}, {key1: "foo", key2: 3}])
+        ensemble.storage.set('dataColumn', Array(15).fill().map((x,i) => ({key1: 'value', key2: i})))
```
evshi, line 5: Can you add an example where we bind data to a mock API with paging params?

evshi (approved):

evshi:
Please add changeset

PR #515, opened 2024-04-08
add humanize method in ensemble.formatter

anserwaseem:
## Describe your changes
Add `humanize` method to ensemble formatter

Use it like `ensemble.formatter.now().humanize(date)`

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/date/dateFormatter.ts
+++ b/packages/framework/src/date/dateFormatter.ts
@@ -5,6 +5,8 @@ import {
 } from "./utils/getPrettyDuration";
 import { getPrettyTime } from "./utils/getPrettyTime";
 import dayjs from "dayjs";
+import duration from "dayjs/plugin/duration";
+import relativeTime from "dayjs/plugin/relativeTime";
 
 export interface EnsembleDateTime {
   getDate: () => string;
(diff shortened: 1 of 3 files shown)


justEhmadSaeed (approved):
lgtm

packages/framework/src/date/dateFormatter.ts
```diff
@@ -60,6 +66,8 @@ export const DateFormatter = (): EnsembleFormatter => {
         dayjs(input).diff(currentDate, "month"),
       getYearsDifference: (input: string): number =>
         dayjs(input).diff(currentDate, "year"),
+      humanize: (input: string): string =>
```
evshi, line 69: I think this should be a bit more descriptive - like humanizeDifferenceFromToday? Or, the other option is to make it composable/chainable with the other smaller functions.

anserwaseem: We'll already be using it composed of `now()` so it's already known that difference is from the current date.
Maybe we should expose same function in `EnsembleFormatter` as well which would take source date as input too. What do you say?

evshi: I think this is probably fine for now. We can always add an optional second param later. Just to clarify when I say compose - I mean allow the user to compose it with other day js functions, not just internally.

evshi (approved):

PR #516, opened 2024-04-09
Set value of Radio as undefined instead of "undefined"

anserwaseem:
## Describe your changes
Fix `value` of Radio being initialized as "undefined"
### Screenshots [Optional]
Before:
image
After:
image

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Form/Radio.tsx
+++ b/packages/runtime/src/widgets/Form/Radio.tsx
@@ -19,7 +19,7 @@ export type RadioWidgetProps = FormInputProps<string> &
   };
 
 export const RadioWidget: React.FC<RadioWidgetProps> = (props) => {
-  const [value, setValue] = useState<string | number>();
+  const [value, setValue] = useState<string | number | undefined>(undefined);
   const { values, rootRef } = useRegisterBindings(
     { ...props, initialValue: props.value, value },
     props.id,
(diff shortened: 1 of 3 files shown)


packages/runtime/src/widgets/Form/Radio.tsx
```diff
@@ -45,7 +45,7 @@ export const RadioWidget: React.FC<RadioWidgetProps> = (props) => {
   const formInstance = Form.useFormInstance();
 
   useEffect(() => {
-    setValue(String(values?.initialValue));
+    setValue(values?.initialValue);
```
sagardspeed2, line 48: @anserwaseem if you remove this type casting here, the initial value, I mean the initial selected radio button will not be selected, you need to click two times to reflect the UI changes.

Please test this also.

sagardspeed2 (changes requested):

packages/runtime/src/widgets/Form/Radio.tsx
sagardspeed2: image
@anserwaseem check this, as radio button has assigned initial value but not rendered on UI

anserwaseem: i missed that, ty.

evshi (approved):

evshi:
@anserwaseem @sagardspeed2 anything outstanding on this one or can we close it out?

sagardspeed2 (approved):

sagardspeed2:
Looking good for me

PR #519, opened 2024-04-10
fix: added support for onSort action handler in datagrid

### LINKED ISSUE #513: Add onSort support for actions to DataColumn
Sometimes you need to sort via API so you need to change your API params when clicking the sort header

**Additional context**

sagardspeed2:
## Describe your changes
Added support for onSort action handler in datagrid widget

Closes #513 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
+++ b/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
@@ -319,10 +319,6 @@ export const DataGrid: React.FC<GridProps> = (props) => {
       case "sort":
         onSortActionCallback(sorter);
         break;
-
-      default:
-        console.log("data grid changes");
-        break;
     }
   };
 
--- a/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
+++ b/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
@@ -304,9 +304,10 @@ export const DataGrid: React.FC<GridProps> = (props) => {
   const onSortActionCallback = useCallback(
     (sorter: SorterResult<unknown>) => {
       if (onSortAction) {
+        console.log({ sorter });
         onSortAction.callback({
-          sort_order: sorter.order,
-          column_title: sorter.column?.title,
+          sortOrder: sorter.order,
+          columnTitle: sorter.column?.title,
         });
       }
     },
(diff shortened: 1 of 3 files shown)


packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -296,12 +298,40 @@ export const DataGrid: React.FC<GridProps> = (props) => {
     return { onChange: handlePageChange, pageSize: values?.pageSize };
   }, [values?.hidePagination, values?.pageSize]);
 
+  const onSortAction = useEnsembleAction(onSort);
+  // page change action
+  const onSortActionCallback = useCallback(
+    (sorter: Partial<TableProps["onChange"]>) => {
+      if (onSortAction) {
+        onSortAction.callback({ sorter });
```
evshi, line 306: What is `sorter`? Can you add an example of how to use this to make an API call with the new sort values?

sagardspeed2: I added example that how we can use sorter to initiate API call

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -296,12 +299,43 @@ export const DataGrid: React.FC<GridProps> = (props) => {
...
   }, [values?.hidePagination, values?.pageSize]);
 
+  const onSortAction = useEnsembleAction(onSort);
+  // page change action
+  const onSortActionCallback = useCallback(
+    (sorter: SorterResult<unknown>) => {
+      if (onSortAction) {
+        onSortAction.callback({
+          sort_order: sorter.order,
+          column_title: sorter.column?.title,
```
evshi, line 309: Can we use camelCase here for consistency? I think we need to pass in the data key for the column being sorted, not just the title.

sagardspeed2:
```js
{
    "sorter": {
        "column": {
            "key": "0",
            "dataIndex": "dummyData",
            "title": "id"
        },
        "order": "ascend",
        "field": "dummyData",
        "columnKey": "0"
    }
}
```

@evshi We are getting this information from sorter, so based on this we can pass data. Now here what you mean by data key property ?

evshi: Each column is bound to a key in the data object. For example if my data object was { foo: 0, bar: "baz" }, then foo and bar would be the keys.

evshi (approved):

PR #521, opened 2024-04-11
fix: added condition to disable free text entry

### LINKED ISSUE #518: Add option to disable free text entry in MultiSelect
Currently users can enter free text into MultiSelect and it becomes a value. Sometimes you just want to limit users to the available values.

sagardspeed2:
## Describe your changes
Added condition to disable free text entry in multiselect

Closes #518 

- [ ] I have added tests

### DIFF
--- a/apps/kitchen-sink/src/ensemble/screens/forms.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/forms.yaml
@@ -99,7 +99,7 @@ View:
                         labelKey: firstName
                         valueKey: email
                         items: ${ensemble.storage.get('dummyData')}
-                        mode: "multiple"
+                        allowCreateOptions: false
                         onItemSelect:
                           executeCode: |
                             console.log('selected');
--- a/apps/kitchen-sink/src/ensemble/screens/forms.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/forms.yaml
@@ -99,6 +99,7 @@ View:
                         labelKey: firstName
                         valueKey: email
                         items: ${ensemble.storage.get('dummyData')}
+                        mode: "multiple"
                         onItemSelect:
                           executeCode: |
                             console.log('selected');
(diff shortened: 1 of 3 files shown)


packages/runtime/src/widgets/Form/MultiSelect.tsx
```diff
@@ -40,6 +40,7 @@ export type MultiSelectProps = {
   items?: Expression<SelectOption[]>;
   onItemSelect?: EnsembleAction;
   hintStyle?: EnsembleWidgetStyles;
+  mode?: "multiple" | "tags";
```
evshi, line 43: Pick a better attribute name - it needs to be intuitive to the end user not just a passthrough of whatever internal component you are using. That is a poor abstraction.

packages/runtime/src/widgets/Form/MultiSelect.tsx
```diff
@@ -127,12 +128,11 @@ const MultiSelect: React.FC<MultiSelectProps> = (props) => {
...
-      values?.options.some((option) =>
-        option.label.toString().toLowerCase().startsWith(value.toLowerCase()),
-      )
-    )
-      setNewOption("");
+    const isOptionExist = values?.options.some((option) =>
+      option.label.toString().toLowerCase().startsWith(value.toLowerCase()),
+    );
+
+    if (isOptionExist || values?.mode !== "tags") setNewOption("");
```
evshi, line 135: Why is this required if the antd mode already handles the new option?

evshi (changes requested):

packages/runtime/src/widgets/Form/MultiSelect.tsx
sagardspeed2: To remove this bottom message when we disable the tags mode, we need this condition

image

apps/kitchen-sink/src/ensemble/screens/forms.yaml
```diff
@@ -99,6 +99,7 @@ View:
                         labelKey: firstName
                         valueKey: email
                         items: ${ensemble.storage.get('dummyData')}
+                        mode: "multiple"
```
evshi, line 102: This needs to be updated.

evshi (approved):

PR #524, opened 2024-04-17
fix: toggle button value change bindings

### LINKED ISSUE #520: ToggleButton binding doesn't update
**Describe the bug**
If you remove the console log, then the following conditional does not update.

https://github.com/atlas-health/atlas-navigator-ui/blob/d78f94444f8b0865d2a2801897b99981864505fe/src/ensemble/common/screens/workflow/widgets/workflows/2_drugFlow/2_2_order.yaml#L30-L33

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

sagardspeed2:
## Describe your changes
Fix toggle button value change bindings

Closes #520 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/ToggleButton.tsx
+++ b/packages/runtime/src/widgets/ToggleButton.tsx
@@ -45,20 +45,16 @@ export type ToggleButtonProps = {
 } & EnsembleWidgetProps<ToggleButtonPropsStyles & EnsembleWidgetStyles>;
 
 export const ToggleButton: React.FC<ToggleButtonProps> = ({
-  id: propId,
+  id,
   value: givenValue,
   items,
   onChange,
   styles,
 }) => {
   const [value, setValue] = useState(givenValue);
-  const { values, rootRef, id } = useRegisterBindings(
-    { value, styles },
-    propId,
-    {
-      setValue,
-    },
-  );
+  const { values, rootRef } = useRegisterBindings({ value, styles }, id, {
+    setValue,
+  });
 
   // onchange action handler
   const action = useEnsembleAction(onChange);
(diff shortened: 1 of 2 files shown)


packages/runtime/src/widgets/ToggleButton.tsx
```diff
@@ -45,47 +45,72 @@ export type ToggleButtonProps = {
...
-
-        action.callback({
-          [id!]: {
-            value: newValue,
-            setValue,
-          },
-        });
-      }
+      action.callback({
+        [id]: {
```
evshi, line 72: I think we just pass the value here to be consistent with other change widgets.

packages/runtime/src/widgets/ToggleButton.tsx
```diff
@@ -45,47 +45,72 @@ export type ToggleButtonProps = {
...
+  );
+
+  // onchange action handler
   const action = useEnsembleAction(onChange);
   const onChangeCallback = useCallback(
-    (_event: React.MouseEvent<HTMLElement>, newValue: string) => {
+    (newValue: string) => {
       if (!action) {
         return;
       }
```
evshi, line 69: So the issue was we don't update the value if there is no action?

evshi (approved):

packages/runtime/src/widgets/ToggleButton.tsx
sagardspeed2: Yes

packages/runtime/src/widgets/ToggleButton.tsx
sagardspeed2: @evshi  Should I remove other consistent values except the value ?

evshi: Yes

PR #534, opened 2024-04-19
fix: added totalRows params in datagrid

sagardspeed2:
## Describe your changes
Added `totalRows` params in datagrid for pagination

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
+++ b/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
@@ -167,6 +167,11 @@ export const DataGrid: React.FC<GridProps> = (props) => {
       setPageSize,
     },
   );
+
+  useEffect(() => {
+    setPageSize(values?.pageSize);
+  }, [values?.pageSize]);
+
   const { namedData } = useTemplateData({ ...itemTemplate });
   const headerStyle = values?.styles?.headerStyle;
   const onTapAction = useEnsembleAction(itemTemplate.template.properties.onTap);
--- a/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
+++ b/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
@@ -6,6 +6,7 @@ import {
   type ResizableState,
 } from "react-resizable";
 import {
+  useEnsembleStorage,
   useTemplateData,
   type Expression,
   type EnsembleWidget,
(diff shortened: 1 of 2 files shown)


sagardspeed2:
@evshi this params is important for pagination to render the number of pages, for example, on the first API call, I just get 10 records through API call and total records are around 100, and I need to show them on dataGrid per page 10, so by this params, I can render the 10 page in pagination when I have only 10 records, and onPage change I can get another 10 records

evshi:
But this can be done by just filling the binding array w/ empty values.

sagardspeed2:
@evshi That is not the right solution, we can just pass the proper params to components and that will do it correctly.

evshi:
@sagardspeed2 why is it not the right solution? Why is this solution better? IMO, this is only a partial solution because it doesn't fit into the solution for the entire use case.

If I am fetching paginated data, I will need a local storage array to concatenate my retrieved set of data anyway. That is source of truth about what is available in the grid. Why do I need extra parameters to describe what my data looks like?

evshi:
In other words, what can you express with this parameter that isn't expressible in your data array already?

sibteali786:
> But this can be done by just filling the binding array w/ empty values.

are those empty values be equal to number of total rows we have ? If that is the case i ll try it.

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -277,28 +287,42 @@ export const DataGrid: React.FC<GridProps> = (props) => {
...
     [onPageChangeAction],
   );
 
+  const storage = useEnsembleStorage();
+
   // handle page change
-  const handlePageChange = (page: number): void => {
-    onPageChangeActionCallback(page);
+  const handlePageChange = (page: number, newPageSize: number): void => {
+    storage.set(resolvedWidgetId, { page, pageSize: newPageSize });
```
evshi, line 306: Let's remove this for now. We will handle the current page indexing and loading state separately.

anserwaseem (approved):

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -135,6 +136,7 @@ export const DataGrid: React.FC<GridProps> = (props) => {
   const [colWidth, setColWidth] = useState<{
     [key: number]: number | undefined;
   }>({});
+  const [curPage, setCurPage] = useState<number>(1);
```
evshi, line 139: Would be nice to have this as a binding also.

evshi (approved):

PR #542, opened 2024-04-23
fix: invalid invocation error

sagardspeed2:
## Describe your changes
fix illegal invocation error

### Screenshots [Optional]
image

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/hooks/useHtmlPassThrough.tsx
+++ b/packages/framework/src/hooks/useHtmlPassThrough.tsx
@@ -1,5 +1,5 @@
 import type { RefCallback } from "react";
-import { mapKeys, isObject } from "lodash-es";
+import { mapKeys, isObject, merge } from "lodash-es";
 import { useCallback } from "react";
 import { error } from "../shared";
 
@@ -11,6 +11,10 @@ export const useHtmlPassThrough = (
     (node: never) => {
       if (node && "setAttribute" in node) {
         if (isObject(htmlAttributes)) {
+          if (testId) {
+            merge(htmlAttributes, { "data-testid": testId });
+          }
+
           const htmlAttributesObj = mapKeys(htmlAttributes, (_, key) =>
             key.toLowerCase(),
           );
@@ -22,14 +26,6 @@ export const useHtmlPassThrough = (
               error(e);
             }
           });
-        }
-
-        if (testId) {
-          try {
-            (node as HTMLElement).setAttribute("data-testid", testId);
-          } catch (e) {
-            error(e);
-          }
         }
       }
     },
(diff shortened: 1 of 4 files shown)


packages/framework/src/hooks/useHtmlPassThrough.tsx
```diff
@@ -23,14 +27,6 @@ export const useHtmlPassThrough = (
...
           });
         }
-
-        if (testId) {
-          try {
-            (node as HTMLElement).setAttribute("data-testid", testId);
-          } catch (e) {
-            error(e);
-          }
-        }
```
evshi, line 33: Why does this fix the issue? It's the same call as what's in the for loop.

sagardspeed2: @evshi honestly I have no idea that why this fix resolved by removing this, I just goes through the code structure and noticed that we no need to use this, as we can manage this with the for loop

anserwaseem: Firefox shows the error more descriptively.
I believe this is the issue with the way we are using rootRef in antd Table. Getting the table containerzied inside a div with ref={rootRef} solves the problem without disturbing useHtmlPassThrough file

image

evshi: Interesting - I wonder why putting it into a single map works. This probably means the ref changes...I think wrapping it in a div is reasonable.

evshi (commented):
@sagardspeed2 you need to check the scroll behavior is not impacted by wrapping the div.

packages/framework/src/hooks/useHtmlPassThrough.tsx
sagardspeed2: @anserwaseem as per my view, no need to remove the changes from useHtmlPassThrough hook, as I just combine the 2 different attributes into one loop.

sagardspeed2:
> @sagardspeed2 you need to check the scroll behavior is not impacted by wrapping the div.

@evshi yup, I tested scroll features on the datagrid and it is working after changing the refs to parent wrapper

evshi (approved):

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -277,7 +277,7 @@ export const DataGrid: React.FC<GridProps> = (props) => {
       const { scrollTop, scrollHeight, clientHeight } = container;
 
       // Check if the user has scrolled to the bottom
-      if (scrollTop + clientHeight === scrollHeight) {
+      if (Math.ceil(scrollTop + clientHeight) >= scrollHeight) {
```
anserwaseem, line 280: please add this fix for onScrollEnd in GridView as well

anserwaseem (changes requested):

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
sagardspeed2: Added

anserwaseem (approved):

PR #553, opened 2024-04-29
feat: added stack widget

### LINKED ISSUE #545: Add Stack widget
[A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]](https://docs.ensembleui.com/widgets/stack)

**Additional context**

sagardspeed2:
## Describe your changes
Added stack widget support

Closes #545 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Stack.tsx
+++ b/packages/runtime/src/widgets/Stack.tsx
@@ -34,68 +34,64 @@ const calculateStackAlignment = (id: string, alignment: string): string => {
   switch (alignment) {
     case "topLeft":
       alignmentStyles = `
-        top: 0;
-        left: 0;
+        align-items: start;
+        justify-items: left;
       `;
       break;
 
     case "topCenter":
       alignmentStyles = `
-        top: 0;
-        left: 50%;
-        transform: translateX(-50%);
+        align-items: start;
+        justify-items: center;
       `;
       break;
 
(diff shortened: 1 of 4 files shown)


packages/runtime/src/widgets/Stack.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+      return {
+        ...child,
+        properties: {
+          ...child.properties,
+          styles: merge(styles, stackStyles),
+        },
+      };
+    });
+  }, [children, values]);
```
evshi, line 139: Rather than use JS to do this I think we can just inject a style tag with a children selector

packages/runtime/src/widgets/Stack.tsx
```diff
@@ -29,124 +28,104 @@ type StackProps = {
...
       break;
 
     case "default":
-      alignmentStyles = {};
+      alignmentStyles = "";
       break;
   }
 
-  return { position: "absolute", ...alignmentStyles };
+  return `#${id} > :not(:first-child) {position: absolute; ${alignmentStyles}}`;
```
evshi, line 108: Why does this not apply to first child?

evshi (commented):
@sagardspeed2 how does this behave if the first child is not the largest item in the stack

evshi (approved):

PR #554, opened 2024-04-29
feat: added support to define custom widget in screen

### LINKED ISSUE #529: Support defining custom widget in screen

sagardspeed2:
## Describe your changes
Added support to define custom widget in screen

Closes #529 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/registry.tsx
+++ b/packages/runtime/src/registry.tsx
@@ -1,19 +1,35 @@
 import { Alert } from "antd";
 import type { ReactElement } from "react";
 
+type WidgetComponent<T> = React.FC<T>;
+
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const registry: Record<string, React.FC<any> | undefined> = {};
+const mainRegistry: Record<string, WidgetComponent<any> | undefined> = {};
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const backupRegistry: Record<string, WidgetComponent<any> | undefined> = {};
 
 export const WidgetRegistry = {
-  register: <T,>(name: string, component: React.FC<T>): void => {
-    registry[name] = component;
+  register: <T,>(name: string, component: WidgetComponent<T>): void => {
+    if (name in mainRegistry) {
+      backupRegistry[name] = mainRegistry[name];
+    }
+
+    mainRegistry[name] = component;
   },
-  find: (name: string): React.FC | ReactElement => {
-    const Widget = registry[name];
+  find: (name: string): WidgetComponent<any> | ReactElement => {
+    const Widget = mainRegistry[name];
     if (!Widget) {
       return <UnknownWidget missingName={name} />;
     }
     return Widget;
+  },
+  unMount: (name: string): void => {
+    if (name in backupRegistry) {
+      mainRegistry[name] = backupRegistry[name];
+      delete backupRegistry[name];
+    } else {
+      delete mainRegistry[name];
+    }
   },
 };
 
--- a/packages/runtime/src/registry.tsx
+++ b/packages/runtime/src/registry.tsx
@@ -23,7 +23,7 @@ export const WidgetRegistry = {
     }
     return Widget;
   },
-  unMount: (name: string): void => {
+  unregister: (name: string): void => {
     if (name in backupRegistry) {
       mainRegistry[name] = backupRegistry[name];
       delete backupRegistry[name];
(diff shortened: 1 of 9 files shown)


packages/runtime/src/runtime/screen.tsx
```diff
@@ -37,6 +40,28 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
...
+      return;
+    }
+
+    // load screen custom widgets
+    screen.customWidgets.forEach((customWidget) => {
+      WidgetRegistry.register(
+        customWidget.name,
+        createCustomWidget(customWidget),
+      );
+    });
```
evshi, line 56: This does not scope the custom widget to the current screen.

evshi (changes requested):

packages/runtime/src/registry.tsx
```diff
@@ -1,20 +1,36 @@
 import { Alert } from "antd";
 import type { ReactElement } from "react";
 
+type WidgetComponent<T> = React.FC<T>;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const mainRegistry: Record<string, WidgetComponent<any> | undefined> = {};
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const registry: Record<string, React.FC<any> | undefined> = {};
+const backupRegistry: Record<string, WidgetComponent<any> | undefined> = {};
```
evshi, line 9: What is the purpose of having two registries?

packages/runtime/src/registry.tsx
```diff
@@ -1,20 +1,36 @@
...
-  find: (name: string): React.FC | ReactElement => {
-    const Widget = registry[name];
+  find: (name: string): WidgetComponent<any> | ReactElement => {
+    const Widget = mainRegistry[name];
     if (!Widget) {
       return <UnknownWidget missingName={name} />;
     }
     return Widget;
   },
+  unMount: (name: string): void => {
```
evshi, line 26: Rename to `unregister`. Functions should be named symmetrically.

packages/runtime/src/registry.tsx
sagardspeed2: When user created any custom widget with the same name of our in-build widgets, that time we can store our in-built widget into backupRegistry and load the custom widget into mainRegistry and serve the widget, and when screen gets destroyed, we remove that custom widget from mainRegistry and pick our in-built widget from the backupRegistry and load into mainRegistry

evshi (commented):
You don't need an entirely separate map - you can just memoize the old implementation and restore it on unmount.

sagardspeed2:
> You don't need an entirely separate map - you can just memoize the old implementation and restore it on unmount.

This is just one object that useful to memoize the old implementation by storing that implementation temporary and then when screen unregister, we again remove the old implementation from the object.

packages/runtime/src/registry.tsx
```diff
@@ -1,20 +1,25 @@
 import { Alert } from "antd";
 import type { ReactElement } from "react";
 
+type WidgetComponent<T> = React.FC<T>;
+
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const registry: Record<string, React.FC<any> | undefined> = {};
+export const registry: { [key: string]: WidgetComponent<any> | undefined } = {};
```
evshi, line 7: Don't export the registry, it is supposed to be private to this module.

packages/runtime/src/runtime/screen.tsx
```diff
@@ -26,6 +28,7 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
 }) => {
   // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
   const { state, search, pathname } = useLocation();
+  const [isLoaded, setIsLoaded] = useState<boolean>(false);
```
evshi, line 31: Can we rename to isInitialized?

packages/runtime/src/runtime/screen.tsx
```diff
@@ -37,6 +40,51 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
...
+    }
+
+    // initial widget values store
+    const initialWidgetValues: {
+      [key: string]: WidgetComponent<any>;
+    } = {};
+
+    // load screen custom widgets
+    screen.customWidgets?.forEach((customWidget) => {
+      const originalImplementation = findWidget(customWidget.name);
```
evshi, line 57: Keep this on WidgetRegistry object. Name it `findOrNull`

evshi (approved):

PR #558, opened 2024-05-02
fix: search in multiselect

### LINKED ISSUE #555: Multiselect search not working
**Describe the bug**

Image

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

sagardspeed2:
## Describe your changes
fix search issue in multiselect

Closes #555 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Form/MultiSelect.tsx
+++ b/packages/runtime/src/widgets/Form/MultiSelect.tsx
@@ -61,35 +61,17 @@ const MultiSelect: React.FC<MultiSelectProps> = (props) => {
     },
   );
 
+  // load initial values
   useEffect(() => {
     if (!selectedValues && isArray(values?.initialValue)) {
       setSelectedValues(values?.initialValue);
     }
   }, [values?.initialValue]);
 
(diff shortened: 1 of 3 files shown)


packages/runtime/src/widgets/Form/MultiSelect.tsx
```diff
@@ -264,6 +268,7 @@ const MultiSelect: React.FC<MultiSelectProps> = (props) => {
             notFoundContent="No Results"
             onChange={handleChange}
             onSearch={handleSearch}
```
evshi, line 270: Why do we need both handleSearch and filter options?

sagardspeed2: image

As per the previous implementation inside the Multiselect widget, when there is no matching option found and user allow to create custom option, we want to show something like this in UI, that's why we need handleSearch

evshi: Can we leave a comment about this?

evshi (approved):

PR #562, opened 2024-05-03
bind invokeAPI name

anserwaseem:

Closes #

- [ ] I have added tests

### DIFF
--- a/.changeset/tricky-pianos-hug.md
+++ b/.changeset/tricky-pianos-hug.md
@@ -1,0 +1,6 @@
+---
+"@ensembleui/react-kitchen-sink": patch
+"@ensembleui/react-runtime": patch
+---
+
+bind invokeAPI name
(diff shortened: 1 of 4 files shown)


sibteali786 (approved):

evshi (changes requested):
You don't need to bind anything to invokeAPI in the code context since it's all javascript already

evshi (commented):
@anserwaseem what is the use case for this? This can create issues with how we create our dependencies since the API name is no longer static. Is this something that flutter supports?

anserwaseem:
@evshi Yes flutter side supports this.

https://github.com/EnsembleUI/ensemble-react/assets/32386713/5841a732-e09a-48e1-9b5a-edc0a9fb0e80

No particular use case for now, i just wanted not to write the same widget again (or use Conditional) just because the api name is different.

evshi:
I see - this will require some special handling. The way our bindings work today we analyze the expressions for the symbols that will become dependencies. invokeAPI.name is just a string. We will need to add handling that if the output of the expression is also a symbol then we need to be dependent on that symbol also.

evshi (approved):
This is probably fine for your use case and it won't break anything.

PR #564, opened 2024-05-06
added curPage in datagrid

### LINKED ISSUE #559: DataGrid has no curPage attribute
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

sagardspeed2:
## Describe your changes
added curPage in dataGrid

Closes #559 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
+++ b/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
@@ -300,6 +300,7 @@ export const DataGrid: React.FC<GridProps> = (props) => {
     const nextPage = newPageSize !== pageSize ? 1 : page;
     setCurPage(nextPage);
     setPageSize(newPageSize);
+    onPageChangeActionCallback(nextPage, newPageSize || 10);
   };
 
   // handle onChange event on table
--- a/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
+++ b/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
@@ -91,6 +91,7 @@ export type GridProps = {
   onSort?: EnsembleAction;
   pageSize?: number;
   totalRows?: number;
+  curPage?: number;
 } & EnsembleWidgetProps<DataGridStyles>;
 
 function djb2Hash(str: string): number {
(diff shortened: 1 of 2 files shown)


evshi (changes requested):
I'm not sure when you added `skipFirstRender` but why is that there?

sagardspeed2:
> I'm not sure when you added `skipFirstRender` but why is that there?

`skipFirstRender` is needed to just void unnecessary API call on the dataGrid widget register, API should only render on the page change or pageSize change

evshi:
Can you please remove? You don't need it.

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -176,7 +180,7 @@ export const DataGrid: React.FC<GridProps> = (props) => {
       setSelectionType,
       setAllowSelection,
       setPageSize,
-      setCurPage,
+      setPage,
```
evshi, line 183: This is an API breaking change by renaming the function. You need to keep the key the same.

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -164,6 +163,11 @@ export const DataGrid: React.FC<GridProps> = (props) => {
     },
   };
 
+  const setPage = (page: number) => {
```
evshi, line 166: Use `useCallback`

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -282,6 +246,55 @@ export const DataGrid: React.FC<GridProps> = (props) => {
...
+    (n: number) => {
+      setCurPage(n);
+      onPageChangeActionCallback(n, pageSize || 10);
+    },
+    [pageSize, onPageChangeActionCallback],
+  );
+
+  // update pageSize callback
+  const updatePageSize = useCallback(
+    (n: number) => {
```
evshi, line 260: nit: rename to better variables.

evshi (approved):

PR #573, opened 2024-05-08
fix: reset options when there are no options on search query

sagardspeed2:
## Describe your changes
Reset options when there are no options on search query

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Search.tsx
+++ b/packages/runtime/src/widgets/Search.tsx
@@ -18,7 +18,7 @@ import type {
 } from "@ensembleui/react-framework";
 import { AutoComplete, Input } from "antd";
 import { SearchOutlined } from "@mui/icons-material";
-import { get, isEmpty, isNull, isNumber } from "lodash-es";
+import { get, isNull, isNumber } from "lodash-es";
 import { WidgetRegistry } from "../registry";
 import type {
   EnsembleWidgetProps,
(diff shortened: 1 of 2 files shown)


packages/runtime/src/widgets/Search.tsx
```diff
@@ -75,10 +75,6 @@ export const Search: React.FC<SearchProps> = ({
   const onSelectAction = useEnsembleAction(onSelect);
 
   useEffect(() => {
-    if (isEmpty(namedData)) {
-      return;
-    }
-
```
sagardspeed2, line 81: @evshi this needs to be remove because of, is there were no records found on specific query, then we need to reset the options list back to empty array.

For example,

https://github.com/EnsembleUI/ensemble-react/assets/31651612/8f846bd1-e44a-44c1-b5e1-3e0083ba1070

evshi: isEmpty checks for `undefined`, `null` and empty array. You still need to check for null/undefined otherwise things will error out below.

sagardspeed2: You are right, but `namedData` will never be null or undefined from the `useTemplateData` hook.

image

evshi (approved):

packages/runtime/src/widgets/Search.tsx
```diff
@@ -121,7 +117,7 @@ export const Search: React.FC<SearchProps> = ({
         onSearchAction.callback({ search: searchValue });
       }
     },
-    onSearch?.debounceMs,
+    onSearch?.debounceMs || 500,
```
evshi, line 120: I think we should keep this 0 by default.

sagardspeed2: okay, let me set it to 0.

PR #574, opened 2024-05-08
Fix enabled property of form items

anserwaseem:
#### Before:
Screenshot 2024-05-09 at 7 24 09 PM
Screenshot 2024-05-09 at 7 23 55 PM

#### After:
image

image

Closes #

- [ ] I have added tests


anserwaseem:
@sagardspeed2 
yes, widget should get disabled when `enabled` is set to false. This PR addresses and corrects the previous reverse behavior.

evshi (approved):
What was the issue before? This seems basically the same expect for the handling of `null`

PR #579, opened 2024-05-13
Fix button icons alignment

anserwaseem:
## Describe your changes
1. Fix button icons alignment
2. Override default styles of Lottie when given from yaml

### Screenshots [Optional]
1.
Before:
image
After:
image

2. for same styles of Lottie:
Before:
image
After:
Screenshot 2024-05-13 at 2 29 01 PM

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/shared/styles.ts
+++ b/packages/runtime/src/shared/styles.ts
@@ -104,13 +104,21 @@ export const getIcon = (name: string): SvgIconComponent | undefined => {
 export const getComponentStyles = (
   name: string,
   styles?: React.CSSProperties,
-  returnAsString = true,
+  options: {
+    importantStyles?:
+      | keyof React.CSSProperties
+      | Array<keyof React.CSSProperties>;
+    returnAsString?: boolean;
+  } = {
+    importantStyles: [],
+    returnAsString: true,
+  },
 ): string | React.CSSProperties => {
   const styleNames = Object.keys(styles || {}).filter((key) =>
     key.startsWith(name),
   );
-  let result = "";
-  let res: React.CSSProperties = {};
+  let stringResult = "";
+  let result: React.CSSProperties = {};
 
   styleNames.forEach((property) => {
     const styleValue = get(styles, property) as string | undefined;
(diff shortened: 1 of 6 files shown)


packages/runtime/src/shared/styles.ts
```diff
@@ -104,13 +104,21 @@ export const getIcon = (name: string): SvgIconComponent | undefined => {
 export const getComponentStyles = (
   name: string,
   styles?: React.CSSProperties,
-  returnAsString = true,
+  options: {
+    importantStyles?:
+      | keyof React.CSSProperties
+      | Array<keyof React.CSSProperties>;
+    returnAsString?: boolean;
```
evshi, line 111: Why are these needed?

packages/runtime/src/widgets/Button.tsx
```diff
@@ -74,7 +74,13 @@ export const Button: React.FC<ButtonProps> = ({ id, onTap, ...rest }) => {
         }}
       >
         {!loading && (
-          <>
+          <div
+            style={{
+              display: "flex",
+              alignItems: "center",
+              justifyContent: "center",
```
evshi, line 81: Hmm this will break how styling will work by inserting another child element. These styles are already on the parent element - did this regress?

packages/runtime/src/shared/styles.ts
anserwaseem: to override these styles of svg
image

packages/runtime/src/widgets/Button.tsx
anserwaseem: Yea. The label and icons are not aligned since this [PR](https://github.com/EnsembleUI/ensemble-react/pull/572) there's a span containerizing the icons and label, we need to align them

evshi: Weird I didn't notice while testing. Not sure how this fixes the issue if that span is still wrapping the icons and label?

packages/runtime/src/shared/styles.ts
evshi: Can you provide more detail about the problem? Which styles specifically?

You should be able to achieve what you're looking for without these extra properties.

PR #583, opened 2024-05-15
wip: websocket support

### LINKED ISSUE #323: Add support for web sockets

**Additional context**

sagardspeed2:
## Describe your changes
Added websocket feature

Closes #323 

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/shared/actions.ts
+++ b/packages/framework/src/shared/actions.ts
@@ -138,6 +138,12 @@ export interface ExecuteActionGroupAction {
   actions: EnsembleAction[];
 }
 
+export type EnsembleActionHookResult =
+  | {
+      callback: (...args: unknown[]) => unknown;
+    }
+  | undefined;
+
 /**
  * @uiType action
  */
--- a/packages/framework/src/shared/actions.ts
+++ b/packages/framework/src/shared/actions.ts
@@ -31,6 +31,19 @@ export interface InvokeAPIAction {
   onError?: EnsembleAction;
 }
 
+export interface ConnectSocketAction {
+  name: string;
+}
+
+export interface SendSocketMessageAction {
+  name: string;
+  message: { [key: string]: unknown };
+}
+
+export interface DisconnectSocketAction {
+  name: string;
+}
+
 export interface NavigateModalScreenStyles {
   position?: "top" | "right" | "bottom" | "left";
   height?: string;
@@ -154,4 +167,7 @@ export type EnsembleAction =
   | { uploadFiles?: UploadFilesAction }
   | { pickFiles?: PickFilesAction }
   | { navigateUrl?: NavigateUrlAction }
-  | { executeActionGroup?: ExecuteActionGroupAction };
+  | { executeActionGroup?: ExecuteActionGroupAction }
+  | { connectSocket?: ConnectSocketAction }
+  | { messageSocket?: SendSocketMessageAction }
+  | { disconnectSocket?: DisconnectSocketAction };
--- a/packages/framework/src/state/screen.ts
+++ b/packages/framework/src/state/screen.ts
@@ -1,8 +1,9 @@
 import { atom, useAtom, useAtomValue } from "jotai";
 import { focusAtom } from "jotai-optics";
-import { clone, isEqual } from "lodash-es";
+import { clone } from "lodash-es";
 import { useCallback } from "react";
-import type { Response } from "../data";
+import isEqual from "react-fast-compare";
+import type { Response, WebSocketConnection } from "../data";
 import type {
   EnsembleAPIModel,
   EnsembleAppModel,
@@ -14,15 +15,15 @@ import type { WidgetState } from "./widget";
 export interface ScreenContextDefinition {
   app?: EnsembleAppModel;
   model?: EnsembleScreenModel;
-  data: { [key: string]: Response | WebSocket | undefined };
-  widgets: { [key: string]: WidgetState | undefined };
+  data: { [key: string]: Response | WebSocketConnection };
+  widgets: { [key: string]: WidgetState };
   inputs?: { [key: string]: unknown };
   [key: string]: unknown;
 }
 
 export interface ScreenContextActions {
   setWidget: (id: string, state: WidgetState) => void;
-  setData: (name: string, response: Response | WebSocket) => void;
+  setData: (name: string, response: Response | WebSocketConnection) => void;
   setCustom: (id: string, data: unknown) => void;
 }
 
@@ -67,7 +68,7 @@ export const useScreenData = (): { apis?: EnsembleAPIModel[] } & {
   const [data, setDataAtom] = useAtom(screenDataAtom);
 
   const setData = useCallback(
-    (name: string, response: Response | WebSocket) => {
+    (name: string, response: Response | WebSocketConnection) => {
       if (isEqual(data[name], response)) {
         return;
       }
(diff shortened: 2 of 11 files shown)


packages/framework/src/state/screen.ts
```diff
@@ -1,28 +1,28 @@
 import { atom, useAtom, useAtomValue } from "jotai";
 import { focusAtom } from "jotai-optics";
-import { clone } from "lodash-es";
+import { clone, isEqual } from "lodash-es";
 import { useCallback } from "react";
-import isEqual from "react-fast-compare";
```
sagardspeed2, line 5: @evshi, we need this change because react-fast-compare can't properly compare objects at deeper levels. See the screenshot where it wrongly returns true even though readyState differs between records.

image

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -299,6 +315,179 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
   return invokeApi;
 };
 
+export const useConnectSocket: EnsembleActionHook<ConnectSocketAction> = (
+  action,
+) => {
+  const { sockets, setData } = useScreenData();
+  const [isComplete, setIsComplete] = useState<boolean>();
+  const [isLoading, setIsLoading] = useState<boolean>();
```
evshi, line 323: Change isComplete to isOpen. Socket lifecycle is different from normal HTTP request

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -299,6 +315,179 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
...
+        ws.onclose = (): void => {
+          onSocketDisconnectAction?.callback();
+        };
+
+        setData(socket.name, ws);
+      } catch (e) {
+        logError(e);
+      } finally {
+        setIsLoading(false);
+        setIsComplete(true);
```
evshi, line 372: These should be set in the event hooks on the socket itself. All this is doing is created a Websocket object but it's not doing the connection.

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -299,6 +315,179 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
...
+    if (!socket || isComplete !== false || isLoading) {
+      return;
+    }
+
+    const fireRequest = (): void => {
+      setIsLoading(true);
+      try {
+        const socketInstance = data[socket.name] as WebSocket;
+        if (socketInstance) {
+          socketInstance.send(JSON.stringify(evaluatedInputs));
```
evshi, line 426: Are these awaitable?

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -299,6 +315,179 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
...
+    if (!socket || isComplete !== false || isLoading) {
+      return;
+    }
+
+    const fireRequest = (): void => {
+      setIsLoading(true);
+      try {
+        const socketInstance = data[socket.name] as WebSocket;
+        if (socketInstance) {
+          socketInstance.close();
```
evshi, line 475: Awaitable?

packages/runtime/src/runtime/websocket.tsx
```diff
@@ -0,0 +1,47 @@
...
+    return;
+  }
+
+  const ws = new WebSocket(socket.uri);
+  screenData.setData(socket.name, ws);
+
+  return ws;
+};
+
+export const handleMessageSocket = (
```
evshi, line 18: Can we reuse these in the hooks?

evshi (changes requested):
Check the awaitables and try to consolidate the code.

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
sagardspeed2: No, because this method add the message in queue of the data need to be transmitted on server through socket. so this will just add the message in queue and return response.

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
sagardspeed2: No, this method will close the socket by closing handshake after checking there is no message left in queue. once the queue will be empty then closing handshake will begin and close the socket

packages/framework/src/shared/actions.ts
```diff
@@ -31,6 +31,19 @@ export interface InvokeAPIAction {
...
+}
+
+export interface SendSocketMessageAction {
+  name: string;
+  message: { [key: string]: unknown };
+}
+
+export interface DisconnectSocketAction {
+  name: string;
+}
```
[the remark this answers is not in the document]
sagardspeed2: I did not understand the part of moving implementation in own file ? as this types will be used in EnsembleAction interface.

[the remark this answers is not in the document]
evshi: They can be imported into the file using `import type`

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -299,6 +311,111 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
...
+  }, []);
+
+  useEffect(() => {
+    if (!action || isComplete !== false) {
+      return;
+    }
+
+    handleDisconnectSocket(screenData, action.name);
+    setIsComplete(true);
+  }, [screenData, action, isComplete]);
```
evshi, line 414: Don't need to track completion state here just return a callback with the handleDisconnect fn.

packages/runtime/src/runtime/websocket.tsx
```diff
@@ -0,0 +1,68 @@
...
+  onOpen?: EnsembleActionHookResult,
+  onMessage?: EnsembleActionHookResult,
+  onClose?: EnsembleActionHookResult,
+): WebSocketConnection | undefined => {
+  const socket = screenData.sockets?.find((model) => model.name === socketName);
+  if (!socket) {
+    return;
+  }
+
+  const ws = new WebSocket(socket.uri);
```
evshi, line 19: You should check if there's an existing socket that is connected before opening a new one.

packages/runtime/src/runtime/websocket.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  ws.onopen = (): void => {
```
evshi, line 21: Don't initialize this function if the callback is null

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -299,6 +311,111 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
   return invokeApi;
 };
 
+export const useConnectSocket: EnsembleActionHook<ConnectSocketAction> = (
+  action,
+) => {
+  const screenData = useScreenData();
+  const [isOpen, setIsOpen] = useState<boolean>();
```
evshi, line 318: You don't need this state variable, just return the callback to connect.

evshi (commented):
Some more cleanup and cases to handle

evshi (approved):

PR #628, opened 2024-05-30
fix: read storage without get

### LINKED ISSUE #623: Support reading storage variable directly without get
Instead of `ensemble.storage.get('attr')`, we should be able to reference attr like so `ensemble.storage.attr`

**Additional context**

sagardspeed2:
## Describe your changes
Added support to read storage value without get method

Closes #623 

- [x] I have added tests

### DIFF
--- a/packages/framework/src/hooks/useEnsembleStorage.ts
+++ b/packages/framework/src/hooks/useEnsembleStorage.ts
@@ -2,27 +2,24 @@ import { useAtom } from "jotai";
 import { createJSONStorage, atomWithStorage } from "jotai/utils";
 import { clone, merge } from "lodash-es";
 import { useMemo } from "react";
+import { type EnsembleStorage } from "../shared/ensemble";
 
-export interface EnsembleStorage {
-  set: (key: string, value: unknown) => void;
-  get: (key: string) => unknown;
-  delete: (key: string) => unknown;
-}
-
-const backingStorage = createJSONStorage<Record<string, unknown>>(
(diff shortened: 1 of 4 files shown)


packages/framework/src/hooks/useEnsembleStorage.ts
```diff
@@ -33,12 +37,16 @@ export const useEnsembleStorage = (): EnsembleStorage => {
     [setStorage, storageBuffer],
   );
 
+  useMemo(() => {
```
evshi, line 40: What is this hook for?

packages/framework/src/hooks/useEnsembleStorage.ts
```diff
@@ -3,26 +3,30 @@ import { createJSONStorage, atomWithStorage } from "jotai/utils";
 import { clone, merge } from "lodash-es";
 import { useMemo } from "react";
 
-export interface EnsembleStorage {
+export type EnsembleStorage = {
```
evshi, line 6: Let's keep the definition in the type file and import it here

sharjeelyunus:
@sagardspeed2 does this also works for set? 
user should also be able to set the values in storage without using set like this: `ensemble.storage.apidata = response.body`

sagardspeed2:
> ser should also be able to set the values in st

No, this doesn't support in react at this time.

evshi (approved):

evshi:
@sharjeelyunus made this to track for now https://github.com/EnsembleUI/ensemble-react/issues/635

PR #640, opened 2024-06-03
fix: convert autocomplete widget with select widget

### LINKED ISSUE #588: Render item-template inside search component value
Sometimes the search key and presentation can differ. We already support rendering options via item-template but when a user selects something only the key gets rendered in the input.

**Additional context**

sagardspeed2:
## Describe your changes
Convert autocomplete widget with select widget

Closes #588 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Search.tsx
+++ b/packages/runtime/src/widgets/Search.tsx
@@ -119,12 +119,23 @@ export const Search: React.FC<SearchProps> = ({
 
   const handleSelect = useCallback(
     (selectedValue: unknown): void => {
-      setValue(selectedValue);
-      if (onSelectAction) {
-        onSelectAction.callback({ selectedValue });
+      if (isObject(itemTemplate) && !isEmpty(evaluatedNamedData.namedData)) {
+        setValue(selectedValue);
+        const selectedOption = evaluatedNamedData.namedData.find((option) => {
+          const optionValue = get(option, [
+            itemTemplate.name,
+            searchKey || "value",
+          ]) as string | number;
+
+          return optionValue === selectedValue;
+        });
+
+        onSelectAction?.callback({
+          value: get(selectedOption, [itemTemplate.name]),
+        });
       }
     },
-    [onSelectAction],
+    [onSelectAction, itemTemplate, evaluatedNamedData, searchKey],
   );
 
   return (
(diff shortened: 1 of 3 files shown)


packages/runtime/src/widgets/Search.tsx
```diff
@@ -40,21 +36,18 @@ export type SearchProps = {
...
   iconStyles?: EnsembleWidgetStyles;
 } & EnsembleWidgetProps &
-  HasItemTemplate;
+  HasItemTemplate & {
+    "item-template"?: { value: Expression<string> };
+  };
 
 export const Search: React.FC<SearchProps> = ({
   "item-template": itemTemplate,
-  searchKey,
```
evshi, line 47: We need to support this until atlas is off this API.

evshi (changes requested):
Can you check this does not create regressions in the atlas app

packages/runtime/src/widgets/Search.tsx
```diff
@@ -63,55 +56,48 @@ export const Search: React.FC<SearchProps> = ({
...
-      return options;
+  // rendered options
+  const renderOptions = useMemo(() => {
+    let dropdownOptions = null;
+
+    if (isObject(itemTemplate) && !isEmpty(evaluatedNamedData.namedData)) {
+      const tempOptions = evaluatedNamedData.namedData.map((item: unknown) => {
+        const optionValue = get(item, [itemTemplate.name, searchKey || ""]) as
+          | string
+          | number;
```
evshi, line 81: Should use item.value here if searchKey doesn't exit?

justEhmadSaeed (approved):
if it's not creating regression, we can merge it forward.

packages/runtime/src/widgets/Search.tsx
```diff
@@ -132,74 +119,57 @@ export const Search: React.FC<SearchProps> = ({
 
   const handleSelect = useCallback(
     (selectedValue: unknown): void => {
-      const selectedData = options.find(
-        ({ value: optionValue }) => optionValue === selectedValue,
-      )?.data;
-      setValue(selectedData);
-      onSelectAction?.callback({ value: selectedData });
+      setValue(selectedValue);
```
justEhmadSaeed, line 122: the key shouldn't be renamed and it should set the complete object instead of a single value.

packages/runtime/src/widgets/Search.tsx
sagardspeed2: @evshi Instead of using item.value we need to use itemTemplate.value, this will helpful us when there is only array of string.

PR #654, opened 2024-06-13
fix: storage regression

### LINKED ISSUE #653: Fix Storage Regression Caused by PR#628
**Describe the bug**
Storage variables are bahaving incorrectly, see videos for more details

Before:

https://github.com/EnsembleUI/ensemble-react/assets/32386713/4f5c4700-145e-4bb7-8004-1e2b8c820ee2

After:

https://github.com/EnsembleUI/ensemble-react/assets/32386713/187db582-b0ce-4b6c-ad05-86e712492a9d

**To Reproduce**
go to `/advocate/home`, then click on any record of datagrid to open workflow screen

**Expected behavior**

**Screenshots**

**Additional context**

sagardspeed2:

Closes #653 

- [ ] I have added tests


packages/framework/src/__tests__/evaluate.test.ts
```diff
@@ -34,38 +36,54 @@ return test.value.concat(bar)
...
+      {
+        ensemble: { storage: storageStore },
+      },
+    );
+  });
+
+  // storage value
+  const { result: updatedStorageValue } = renderHook(() =>
+    useAtomValue(storageAtom),
   );
```
evshi, line 61: You shouldn't need this you should just use result.current.value

packages/framework/src/hooks/useEnsembleStorage.ts
```diff
@@ -1,61 +1,48 @@
...
 
-const backingStorage = createJSONStorage<{ [key: string]: unknown }>(
-  () => sessionStorage,
-);
+export interface EnsembleStorage {
+  [key: string]: unknown;
+}
 
-export const screenStorageAtom = atomWithStorage<{ [key: string]: unknown }>(
+export const storageAtom = atomWithSessionStorage<EnsembleStorage>(
```
evshi, line 10: Why rename this? Seems to create a bunch of unnecessary changes.

packages/framework/src/hooks/useScreenContext.tsx
```diff
@@ -80,6 +81,7 @@ const HydrateAtoms: React.FC<
       [appAtom, appContext],
       [themeAtom, themeScope.theme],
       [userAtom, appContext.user],
+      [storageAtom, appContext.storage],
```
evshi, line 84: Where does appContext.storage get initialized?

sagardspeed2: image

inside `useApplicationContext`

packages/framework/src/hooks/useScreenContext.tsx
```diff
@@ -80,6 +81,7 @@ const HydrateAtoms: React.FC<
       [appAtom, appContext],
       [themeAtom, themeScope.theme],
       [userAtom, appContext.user],
+      [screenStorageAtom, appContext.storage],
```
evshi, line 84: How does it behave without this? This technically will only be the storage value when the apps is initialized right? Not the latest storage value?

sagardspeed2: Yes, and we need this because of we need get and set method on app `initialized` to handle the onLoad actions and widgets text renders. if storage is accessed through `get` method.

evshi: This needs to be the state of the storage variable when this screen is rendered, not when the application was loaded.

evshi (commented):
Please fix merge conflicts and test against latest version of atlas

PR #668, opened 2024-06-25
fix: added support for multiple expression evaluation

sagardspeed2:
## Describe your changes
Added support for multiple expression evaluation

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/shared/common.ts
+++ b/packages/framework/src/shared/common.ts
@@ -33,7 +33,7 @@ export const findExpressions = (
   Object.entries(obj).forEach(([key, value]) => {
     const curPath = path.concat(key);
     if (isExpression(value)) {
-      expressionMap.push([curPath.join("."), `\${\`${value || ""}\`}`]);
+      expressionMap.push([curPath.join("."), combineExpressions(value)]);
     } else if (isObject(value)) {
       findExpressions(value, curPath, expressionMap);
     }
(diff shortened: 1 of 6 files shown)


packages/framework/src/shared/common.ts
```diff
@@ -33,7 +33,7 @@ export const findExpressions = (
   Object.entries(obj).forEach(([key, value]) => {
     const curPath = path.concat(key);
     if (isExpression(value)) {
-      expressionMap.push([curPath.join("."), combineExpressions(value)]);
+      expressionMap.push([curPath.join("."), `\${\`${value || ""}\`}`]);
```
evshi, line 36: You only interpolate as string if it's a multiple expression, otherwise if this returns a number or object you will create issues.

packages/framework/src/shared/common.ts
```diff
@@ -47,3 +51,26 @@ export const error = (value: unknown): void => {
...
+      resultParts.push(part.slice(2, -1));
+    } else if (part) {
+      // If the part is a literal, wrap it in single quotes and add to resultParts
+      resultParts.push(`'${part}'`);
+    }
+  });
+
+  // Combine all parts with ' + ' and return the result
+  return resultParts.length ? `\${${resultParts.join(" + ")}}` : input;
+};
```
sagardspeed2, line 76: @evshi I think this function is more good approach than string interpolation. I understand that this is big code compare to string interpolation in one line, but this function in capable to handle all use cases of expression uses into YAML, also can check for single or multiple expression and handle in same way. also it will not change the values type that string interpolation can change.

For example, in string interpolation, if the value is in integer format, it will give back as string, but using this one, it will just return integer.

I also added all use cases example on home page and also added test cases for cover all use cases

packages/framework/src/shared/common.ts
```diff
@@ -4,10 +4,40 @@ export type Expression<T> = string | T;
...
+): maybeExpression is string => {
+  if (!isString(maybeExpression)) {
+    return false;
+  }
+
+  const regex = /\$\{[^}]+\}/g;
+  return regex.test(maybeExpression);
+};
+
+export const isMultipleExpressions = (input: string): boolean => {
```
evshi, line 16: Add some tests for this.

evshi (commented):
Are there tests that assert on the evaluated value also? Since this is a multi-layered problem, we need tests that assert:

1. multiple expression detection works correctly
2. expressions are detected correctly (already exists)
3. evaluated output typing is correct (maybe exists?, if not add)

evshi (approved):

PR #675, opened 2024-06-28
fix: theme definition disparity with flutter

### LINKED ISSUE #646: Disparity between Flutter version for Theme definition; Defining Styles under Theme not working
**Describe the bug**
When defining the app Theme, defining a theme under "Themes:", and putting Styles under that theme does not work, the styles are not reflected in the app, whereas if you define Styles on its own, the styles are reflected in the app.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

sagardspeed2:
## Describe your changes
fix theme definition disparity with flutter

Closes #646 

- [ ] I have added tests


packages/framework/src/parser.ts
```diff
@@ -433,29 +427,44 @@ const unwrapTheme = (theme?: string): EnsembleThemeModel | undefined => {
...
 
-  if (!workingTheme.Styles) {
-    return workingTheme;
+  if (isEmpty(workingTheme.Themes)) {
+    return;
   }
 
-  const expressionMap: string[][] = [];
-  findExpressions(workingTheme.Styles, [], expressionMap);
+  const themes = workingTheme.Themes?.reduce(
```
evshi, line 434: Is this backwards compatible?

evshi (commented):
Can you add some tests here? Specifically on the parser. It should work for both single and multiple themes to avoid regressing atlas. cc @anserwaseem

sagardspeed2:
@evshi I will add test cases for parser, but I think we do not need to add feature that can accept multiple theme files as it is not matching with the flutter theme support.
Also atlas is using only one theme file, so in atlas we need to make very small change related Themes and it will not create any regression

evshi (commented):
Let's add some tests for the parser for the new theme file format and the old one to assert backwards compatibility. Agreed we don't need multiple theme file support

evshi:
@sagardspeed2 can you add the test here?

evshi (approved):

PR #677, opened 2024-07-01
fix: added flutter style hex code support

### LINKED ISSUE #642: Support flutter style hex colors, e.g. 0xFFFFFFFF
Flutter supports rgba in hex format, which is incompatible with web by default. We need to translate these colors to web standard RGBA

**Additional context**

sagardspeed2:
## Describe your changes
Added support for flutter style hex color code support

Closes #642 

- [ ] I have added tests


packages/framework/src/hooks/useEvaluate.ts
```diff
@@ -63,5 +64,18 @@ export const useEvaluate = <T extends Record<string, unknown>>(
...
+    const hexCodesMaps: string[][] = [];
+    findHexCodes(values, [], hexCodesMaps);
+
+    const result = {};
+    hexCodesMaps.forEach(([name, value]) => {
+      set(result, name, value);
+    });
+
+    return result;
+  }, [values]);
```
evshi, line 78: This is only going to work if the hexcode is hardcoded but not if its defined in a theme file or comes from the result of an expression. I think you should do the conversion after the bindings have been evaluated and only do it on the styles object.

evshi (commented):
Can we add tests for expressions and non expressions for this?

evshi:
@sagardspeed2 can we add tests here?

evshi (approved):

PR #690, opened 2024-07-11
fix: support device object

### LINKED ISSUE #676: Support device object like flutter
The device object provides some utility to ensemble devs, react should support the same spec

**Additional context**
https://docs.ensembleui.com/device-object#devicewidth-and-deviceheight

sagardspeed2:
## Describe your changes
added support for device object

added propertied like, platform, width and height

Closes #676 

- [ ] I have added tests

### DIFF
--- a/apps/kitchen-sink/src/ensemble/screens/home.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/home.yaml
@@ -4,6 +4,7 @@ View:
       ensemble.storage.set('yyy' , 'page')
       ensemble.storage.set('zzz' , 'button')
       ensemble.storage.set('aaa' , '1')
+      console.log(device)
 
   header:
     title:
--- a/apps/kitchen-sink/src/ensemble/screens/home.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/home.yaml
@@ -22,6 +22,10 @@ View:
         - Text:
             styles:
               names: heading-1
+            text: ${device.width}
+        - Text:
+            styles:
+              names: heading-1
               color: ${colors.primary['900']}
             text: r@kitchenSink
         - Markdown:
--- a/packages/framework/src/evaluate/binding.ts
+++ b/packages/framework/src/evaluate/binding.ts
@@ -86,7 +86,6 @@ export const createBindingAtom = <T = unknown>(
           languages: appData.application?.languages,
         },
         env: get(envAtom),
-        device: get(deviceAtom),
       },
       screenContext: {
         inputs: get(screenInputAtom),
@@ -107,7 +106,7 @@ export const createBindingAtom = <T = unknown>(
           : undefined,
         formatter: DateFormatter(),
       },
-      context,
+      context: { ...context, device: get(deviceAtom) },
     });
 
     try {
--- a/packages/framework/src/evaluate/binding.ts
+++ b/packages/framework/src/evaluate/binding.ts
@@ -106,7 +106,12 @@ export const createBindingAtom = <T = unknown>(
           : undefined,
         formatter: DateFormatter(),
       },
-      context: { ...context, device: get(deviceAtom) },
+      context: {
+        ...(rawJsExpression.includes("device")
+          ? { device: get(deviceAtom) }
+          : {}),
+        ...context,
+      },
     });
 
     try {
--- a/packages/framework/src/evaluate/binding.ts
+++ b/packages/framework/src/evaluate/binding.ts
@@ -20,6 +20,7 @@ import {
   screenImportScriptAtom,
   userAtom,
   appAtom,
+  deviceAtom,
 } from "../state";
 import { evaluate } from "./evaluate";
 import { createEvaluationContext } from "./context";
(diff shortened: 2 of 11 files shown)


packages/framework/src/hooks/useScreenContext.tsx
```diff
@@ -54,6 +55,7 @@ export const ScreenContextProvider: React.FC<ScreenContextProviderProps> = ({
               inputs: Object.fromEntries(
                 location.searchParams?.entries() ?? [],
               ),
+              device: useAtomValue(deviceAtom),
```
evshi, line 58: Only include the device atom if expression references it.

packages/runtime/src/EnsembleApp.tsx
```diff
@@ -111,6 +113,14 @@ export const EnsembleApp: React.FC<EnsembleAppProps> = ({
     return null;
   }
 
+  const resizeObserver = new ResizeObserver((_) => {
+    updateDeviceData({
+      width: window.innerWidth,
+      height: window.innerHeight,
+    });
+  });
```
evshi, line 121: How is this with performance? This might need to be debounced.

packages/framework/src/state/device.ts
```diff
@@ -0,0 +1,57 @@
...
+export interface DeviceDefinition {
+  platform: string;
+  width: number;
+  height: number;
+  [key: string]: unknown;
+}
+
+export const defaultDevice = {
+  platform:
+    (navigator as Navigator & NavigatorUA).userAgentData?.platform || "",
```
evshi, line 35: What's with the extra typing?

sagardspeed2: `Navigator` type is not comes with `userAgentData`, so need to add `userAgentData` type to avoid lint errors

evshi (commented):
Can you please add an example in kitchen sink where you use device.width in an expression and the value changes?

packages/framework/src/hooks/useScreenContext.tsx
```diff
@@ -80,13 +83,20 @@ const HydrateAtoms: React.FC<
...
     ],
     {
       dangerouslyForceHydrate: true,
     },
   );
 
-  return <>{children}</>;
+  return (
+    <>
+      {" "}
```
evshi, line 95: Remove?

packages/framework/src/hooks/useScreenContext.tsx
```diff
@@ -114,3 +124,29 @@ export const useScreenContext = ():
   );
   return { ...screenContext, setData };
 };
+
+const DeviceObserver = (): null => {
```
evshi, line 128: This doesn't need to be a component it can just be a hook

packages/framework/src/evaluate/context.ts
```diff
@@ -40,5 +44,7 @@ export const createEvaluationContext = ({
   const app = {
     languages: applicationContext.application?.languages,
   };
-  return merge({}, { app, ensemble }, appInputs, screenInputs, context);
+  return merge({}, { app, ensemble }, appInputs, screenInputs, context, {
+    device: applicationContext.device,
```
evshi, line 48: Can we pass this in instead of adding it here?

apps/kitchen-sink/src/ensemble/screens/home.yaml
```diff
@@ -19,6 +19,10 @@ View:
       styles:
         names: ${ensemble.storage.get('yyy')}
       children:
+        - Text:
+            styles:
+              names: heading-1
+            text: ${device.width}
```
evshi, line 25: Does this change as the screen resizes?

packages/framework/src/evaluate/context.ts
sagardspeed2: where should need to pass ?

packages/framework/src/hooks/useScreenContext.tsx
```diff
@@ -114,3 +121,29 @@ export const useScreenContext = ():
   );
   return { ...screenContext, setData };
 };
+
+const useDeviceObserver = (): null => {
```
evshi, line 125: This should be void return type, and you don't need a return statement.

packages/framework/src/evaluate/context.ts
evshi: By the context object - why did you add this here in the first place?

packages/framework/src/hooks/useScreenContext.tsx
```diff
@@ -114,3 +121,29 @@ export const useScreenContext = ():
...
+  const { setData: updateDeviceData } = useDeviceData();
+
+  const debouncedUpdateDeviceData = useMemo(() => {
+    const handleResize = (): void => {
+      updateDeviceData({
+        width: window.innerWidth,
+        height: window.innerHeight,
+      });
+    };
+    return debounce(handleResize, 500);
```
evshi, line 135: Can we reduce the debounce value to 100ms?

packages/framework/src/evaluate/binding.ts
```diff
@@ -105,7 +106,7 @@ export const createBindingAtom = <T = unknown>(
           : undefined,
         formatter: DateFormatter(),
       },
-      context,
+      context: { ...context, device: get(deviceAtom) },
```
evshi, line 109: Can we only include this if device is part of the expression to avoid creating extra dependencies? It should come before context also

evshi (approved):

PR #694, opened 2024-07-12
fix: designated screen load on route

sagardspeed2:
## Describe your changes
fix designated screen load on route

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/entry.tsx
+++ b/packages/runtime/src/runtime/entry.tsx
@@ -6,6 +6,7 @@ import { Outlet, useNavigate, useLocation } from "react-router-dom";
 import { useEffect } from "react";
 import { SideBarMenu } from "./menu";
 import { EnsembleScreen } from "./screen";
+import { keys } from "lodash-es";
 
 interface EnsembleEntryProps {
   entry: EnsembleEntryPoint;
@@ -20,24 +21,32 @@ export const EnsembleEntry: React.FC<EnsembleEntryProps> = ({
 
   const hasMenu = "items" in entry;
   useEffect(() => {
-    if (screen && location.pathname !== `/${screen.name.toLowerCase()}`) {
-      navigate({
-        pathname: `/${screen.name.toLowerCase()}`,
-        search: location.search,
-      });
+    if (!keys(entry).length) {
       return;
     }
 
-    if (!(hasMenu && location.pathname === "/")) {
+    const navigateToPath = (pathname: string, search = ""): void => {
+      navigate({ pathname, search });
+    };
+
+    const getLowerCasePath = (name: string): string => `/${name.toLowerCase()}`;
+
+    const currentPath = location.pathname;
+
+    if (screen && currentPath !== getLowerCasePath(screen.name)) {
+      navigateToPath(getLowerCasePath(screen.name), location.search);
       return;
     }
 
-    const selectedItem =
-      entry.items.find((item) => item.selected) ?? entry.items[0];
-    navigate({
-      pathname: `/${selectedItem.page.toLowerCase()}`,
-      search: location.search,
-    });
+    if (hasMenu) {
+      if (currentPath === "/") {
+        const selectedItem =
+          entry.items.find((item) => item.selected) ?? entry.items[0];
+        navigateToPath(getLowerCasePath(selectedItem.page), location.search);
+      }
+    } else if (currentPath === "/") {
+      navigateToPath(getLowerCasePath(entry.name));
+    }
   }, [entry, hasMenu, navigate, location, screen]);
 
   if (hasMenu) {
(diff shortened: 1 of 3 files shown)


packages/runtime/src/runtime/entry.tsx
```diff
@@ -20,24 +21,32 @@ export const EnsembleEntry: React.FC<EnsembleEntryProps> = ({
...
-    if (!(hasMenu && location.pathname === "/")) {
+    const navigateToPath = (pathname: string, search = ""): void => {
+      navigate({ pathname, search });
+    };
+
+    const getLowerCasePath = (name: string): string => `/${name.toLowerCase()}`;
+
+    const currentPath = location.pathname;
+
+    if (screen && currentPath !== getLowerCasePath(screen.name)) {
```
evshi, line 36: This is my poor naming, can we rename this to initialScreen?

packages/runtime/src/runtime/entry.tsx
```diff
@@ -20,24 +21,32 @@ export const EnsembleEntry: React.FC<EnsembleEntryProps> = ({
...
-    });
+    if (hasMenu) {
+      if (currentPath === "/") {
+        const selectedItem =
+          entry.items.find((item) => item.selected) ?? entry.items[0];
+        navigateToPath(getLowerCasePath(selectedItem.page), location.search);
+      }
+    } else if (currentPath === "/") {
+      navigateToPath(getLowerCasePath(entry.name));
+    }
```
evshi, line 49: I think you keep the check for selectedItem && menu, and then lastly navigate to home...so the priority should be like

if (path is not root)

if (screen)

if (hasMenu && selectedItem)

goHome

it's not really elseif it's just a series of conditions and can return early if met.

packages/runtime/src/runtime/entry.tsx
```diff
@@ -20,24 +21,35 @@ export const EnsembleEntry: React.FC<EnsembleEntryProps> = ({
...
+    };
+
+    const getLowerCasePath = (name: string): string => `/${name.toLowerCase()}`;
+
+    const initialScreen = location.pathname;
+
+    if (initialScreen !== "/") {
+      return;
+    }
+
```
evshi, line 39: This is incorrect. We should navigate to the path specified in the URL

evshi (changes requested):

packages/runtime/src/runtime/entry.tsx
sagardspeed2: You mean `location.pathname` ?

packages/runtime/src/runtime/entry.tsx
```diff
@@ -20,24 +21,35 @@ export const EnsembleEntry: React.FC<EnsembleEntryProps> = ({
...
-      navigate({
-        pathname: `/${screen.name.toLowerCase()}`,
-        search: location.search,
-      });
+    if (!keys(entry).length) {
       return;
     }
 
-    if (!(hasMenu && location.pathname === "/")) {
+    const navigateToPath = (pathname: string, search = ""): void => {
```
evshi, line 28: Can we remove this? it isn't doing much...

packages/runtime/src/runtime/entry.tsx
```diff
@@ -20,24 +21,35 @@ export const EnsembleEntry: React.FC<EnsembleEntryProps> = ({
...
+      navigate({ pathname, search });
+    };
+
+    const getLowerCasePath = (name: string): string => `/${name.toLowerCase()}`;
```
evshi, line 32: We probably don't need this - you can just store this in a variable.

packages/runtime/src/runtime/entry.tsx
evshi: Can you leave a comment why this works?

packages/runtime/src/runtime/entry.tsx
```diff
@@ -20,24 +21,35 @@ export const EnsembleEntry: React.FC<EnsembleEntryProps> = ({
...
+
+    const initialScreen = location.pathname;
```
evshi, line 34: Rename `initialScreen` to `path`
Rename `screen` to `initialScreen`

packages/runtime/src/runtime/entry.tsx
```diff
@@ -20,24 +21,35 @@ export const EnsembleEntry: React.FC<EnsembleEntryProps> = ({
...
       return;
     }
 
-    const selectedItem =
-      entry.items.find((item) => item.selected) ?? entry.items[0];
-    navigate({
-      pathname: `/${selectedItem.page.toLowerCase()}`,
-      search: location.search,
-    });
+    navigateToPath(getLowerCasePath(entry.name));
```
evshi, line 52: Should include search here?

packages/runtime/src/runtime/entry.tsx
```diff
@@ -20,24 +21,35 @@ export const EnsembleEntry: React.FC<EnsembleEntryProps> = ({
 
   const hasMenu = "items" in entry;
   useEffect(() => {
-    if (screen && location.pathname !== `/${screen.name.toLowerCase()}`) {
-      navigate({
-        pathname: `/${screen.name.toLowerCase()}`,
-        search: location.search,
-      });
+    if (!keys(entry).length) {
```
evshi, line 24: Why is this check needed?

evshi (approved):
Can you please sanity check atlas app as well?

PR #701, opened 2024-07-17
fix: added translation support in studio

sagardspeed2:
## Describe your changes
added translation support in studio

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/loader.ts
+++ b/packages/framework/src/loader.ts
@@ -13,7 +13,41 @@ import type {
   ScriptDTO,
   ThemeDTO,
   WidgetDTO,
+  LanguageDTO,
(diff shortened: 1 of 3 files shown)


packages/framework/src/loader.ts
```diff
@@ -13,15 +13,50 @@ import type {
   ScriptDTO,
   ThemeDTO,
   WidgetDTO,
+  LanguageDTO,
 } from "./shared/dto";
 
+export const languageMap: { [key: string]: string } = {
```
evshi, line 19: Can we put this under the i18n folder?

evshi (approved):

PR #702, opened 2024-07-17
fix: props evaluate in flex widget

sagardspeed2:
## Describe your changes
Props evaluation in flex widget

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Flex.tsx
+++ b/packages/runtime/src/widgets/Flex.tsx
@@ -1,4 +1,7 @@
-import { type EnsembleWidget } from "@ensembleui/react-framework";
+import {
+  useRegisterBindings,
+  type EnsembleWidget,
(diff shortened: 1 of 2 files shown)


packages/runtime/src/widgets/Flex.tsx
```diff
@@ -21,9 +24,11 @@ export type FlexProps = {
   EnsembleWidgetProps<FlexStyles & FlexboxStyles>;
 
 export const FlexWidget: React.FC<FlexProps> = (props) => {
+  const { values } = useRegisterBindings({ ...props }, props.id);
```
evshi, line 27: Only pass in the props that belong to the Flex widget here, otherwise it may create issues with the nested widgets.

evshi (approved):

PR #718, opened 2024-07-19
fix: added support of env in preview and studio

### LINKED ISSUE #714: Env Variables are not working
image

sagardspeed2:
## Describe your changes
Added support of env in preview and studio

Closes #714 

- [ ] I have added tests


packages/framework/src/loader.ts
```diff
@@ -55,6 +58,12 @@ const getArtifacts = async (
         languageCode: document.name as string,
         content: document.content as string,
       });
+    } else if (document.type === "config") {
+      config = stringify({
```
evshi, line 62: Why are we stringifying this?

sagardspeed2: Because we are passing yaml into parser and yaml must be in string format, then parser will convert that into json. so I need to convert this json into yaml first and then pass into parser, so we do not need to change anything related parser

evshi: So the config is a JSON object? You don't need to parse it at all then you can just create an object from it.

evshi (approved):

PR #755, opened 2024-07-29
fix: flutter theme disparity issue

sagardspeed2:
## Describe your changes
flutter theme disparity issue

Closes #

- [x] I have added tests


evshi (changes requested):

packages/framework/src/parser.ts
```diff
@@ -108,12 +104,18 @@ export const EnsembleParser = {
...
-        themes[item.name || item.id] = unwrapTheme(clone(item.content));
+    let themes: { [key: string]: EnsembleThemeModel } = {};
+
+    if (app.themes && !isEmpty(app.themes)) {
+      app.themes.forEach((theme) => {
+        const defaultTheme = unwrapTheme(theme.content) as {
+          default: EnsembleThemeModel;
+        };
+
+        themes[theme.name || "default"] = defaultTheme.default;
```
evshi, line 115: There should only ever be one theme file?

evshi: This should be moved into unwrapTheme (handling of default theme)

packages/framework/src/parser.ts
```diff
@@ -438,29 +441,22 @@ const unwrapTheme = (theme?: string): EnsembleThemeModel | undefined => {
     return;
   }
 
-  if (!workingTheme.Styles) {
-    return workingTheme;
+  if (isEmpty(workingTheme.Themes)) {
+    workingTheme.default = Object.assign({}, workingTheme);
+    workingTheme.Themes = ["default"];
```
evshi, line 446: Leave a comment here about how this handles the single theme case.

packages/framework/src/shared/models.ts
```diff
@@ -129,6 +130,7 @@ export interface EnsembleThemeModel {
     } & { [key: string]: string };
   };
   Styles?: { [key: string]: CSSProperties };
+  [key: string]: unknown;
```
evshi, line 133: Please fix this typing to reflect the proper structure.

packages/framework/src/evaluate/binding.ts
```diff
@@ -84,7 +85,9 @@ export const createBindingAtom = <T = unknown>(
         application: {
           theme: get(themeAtom),
           languages: appData.application?.languages,
+          themes: appData.application?.themes,
         },
+        themeName: get(selectedThemeNameAtom),
```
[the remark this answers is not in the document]
sagardspeed2: Here, 

`theme` is selected theme JSON,
`themes` is the array of multiple available theme names
`themeName` is the selected theme name

as per the ensemble documentation,
we need two variables, `theme` -> selected theme name, and `themes` -> available theme names array

if you see our context.ts, I am exactly using this variables as per the documentation and flutter support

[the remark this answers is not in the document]
sagardspeed2: image

packages/framework/src/shared/models.ts
sagardspeed2: Hey, we need this. I added comment there to explain the usecase of this,
actually in theme file, there will be multiple themes like, Dark, Primary, Light and so on, so this is for that usecase

evshi: Yes I understand, but this can be greatly improved since you know the structure of the theme - create the rest of the type not just make it `unknown`

packages/framework/src/evaluate/binding.ts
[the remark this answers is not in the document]
evshi: Can we just consolidate `theme` and `themeName` into a single variable and rename to `selectedTheme`?

packages/framework/src/hooks/useApplicationContext.tsx
```diff
@@ -68,11 +67,14 @@ const HydrateAtoms: React.FC<
     appContext: ApplicationContextDefinition;
   }>
 > = ({ appContext, children }) => {
-  const activeThemeName = useAtomValue(selectedThemeNameAtom);
+  const activeThemeName =
+    useAtomValue(themeAtom)?.name || keys(appContext.application?.themes)[0];
```
evshi, line 71: `useAtomValue` should be on it's own line per react hooks convention.

packages/framework/src/hooks/useThemeContext.tsx
```diff
@@ -13,26 +13,21 @@ export const CustomThemeContext = createContext<CustomTheme>({});
 
 export const useThemeScope = (): {
   theme: EnsembleThemeModel | undefined;
-  themeName: string;
+  themeName?: string;
```
evshi, line 16: Why is this optional? Does flutter return null or default if there is a single theme?

packages/framework/src/parser.ts
```diff
@@ -59,6 +56,10 @@ export interface EnsembleWidgetYAML {
   };
 }
 
+export type EnsembleThemeParser = {
```
evshi, line 59: Rename `EnsembleThemeYAML` for consistency.

packages/framework/src/parser.ts
```diff
@@ -137,7 +138,7 @@ export const EnsembleParser = {
         menu ||
         (screens as EnsembleScreenModel[]).find((screen) => screen.isRoot) ||
         screens[0],
-      themes,
+      themes: themes || { default: defaultThemeDefinition },
```
evshi, line 141: Shouldn't this be an array?

sagardspeed2: No, this will be object, so we can select any selected theme from themes using theme name

evshi (approved):
Please check this against atlas app

PR #761, opened 2024-07-31
707 support mock api

### LINKED ISSUE #707: Support Mock API
https://docs.ensembleui.com/apis/use-mock-api-response

jkorn8:
## Describe your changes
- Added support for a mockResponse field in APIs
- Successfully added evaluation for this mockResponse field
- Added a toggleable useMockResponse variable (In progress)

#707 

Closes #707 

- [✅] I have performed a self-review of my code
- [✅] I have added tests
- [✅] I have added a changeset `pnpm changeset add`
- [✅] I have added example usage in the kitchen sink app


packages/framework/src/evaluate/binding.ts
```diff
@@ -88,6 +89,7 @@ export const createBindingAtom = <T = unknown>(
         },
         env: get(envAtom),
         secrets: get(secretAtom),
+        useMockResponse: useMockResponse
```
evshi, line 92: This is likely unneeded since this is primary for read expressions but for completeness you will need to create a read-only atom with the value and call get on the atom here so the binding automatically updates.

packages/framework/src/hooks/useMockResponse.ts
```diff
@@ -0,0 +1,5 @@
+export let useMockResponse: boolean = false;
```
evshi, line 1: Like we discussed this should end up being a map of appIds -> config object (that extends the app model) with the useMockResponse attribute in it. Then you can pass the object around and use assignment on it.

Think of this map as a global store of app configuration.

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -324,6 +337,56 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
...
+            ...context,
+            response: mockResponse,
+          });
+        } else {
+          onAPIErrorAction?.callback({ ...context, error: mockResponse });
+          onInvokeAPIErrorAction?.callback({ ...context, error: mockResponse });
+        }
+
+        setIsComplete(true);
+        setIsLoading(false);
```
evshi, line 385: Let's put all the mock handling logic into a separate function in a new `mock` module. You should be able to call the utility from the fetcher as well.

packages/runtime/src/runtime/mock.tsx
```diff
@@ -0,0 +1,37 @@
...
+            "Improperly formatted mock response: Malformed mockResponse object",
+        );
+
+    if (
+        !evaluatedMockResponse.response.statusCode ||
+        typeof evaluatedMockResponse.response.statusCode !== "number"
+    )
+        throw new Error(
+            "Improperly formatted mock response: Incorrect Status Code. Please check that you have included a status code and that it is a number",
+        );
```
evshi, line 25: You can use `isEmpty` and `isNumber` from lodash for validations here.

packages/framework/src/hooks/useMockResponse.ts
```diff
@@ -0,0 +1,13 @@
+const mockResponseTable: Record<string, boolean> = {};
```
evshi, line 1: Yeah let's make this a map from string to object, with the useMockResponse as the first attribute.

packages/framework/src/hooks/useMockResponse.ts
```diff
@@ -0,0 +1,13 @@
...
+    if (typeof appId == 'undefined' || !(appId in mockResponseTable))
+        return false;
+    return mockResponseTable[appId];
+}
+
+export const setUseMockResponse = (appId: string | undefined, value: boolean) => {
+    if (typeof appId == 'undefined')
+        return;
+    mockResponseTable[appId] = value
+};
```
evshi, line 13: Not sure if these functions have much value - you should be able to modify the object directly.

packages/framework/src/evaluate/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from "./evaluate";
 export * from "./binding";
 export * from "./context";
+export * from "./useMockResponse";
```
evshi, line 4: Can we move this into it's own file outside of evaluate folder?

packages/framework/src/evaluate/useMockResponse.ts
```diff
@@ -0,0 +1,18 @@
+const appConfigTable: { [key: string]: AppConfig } = {};
+
+interface AppConfig {
+  useMockResponse: boolean;
+}
+
+export const isUsingMockResponse = (appId: string | undefined): boolean => {
+  if (typeof appId === "undefined" || !(appId in appConfigTable)) return false;
```
evshi, line 8: You can use just `!appId` which will check for both null and undefined.

packages/framework/src/hooks/useScreenData.ts
```diff
@@ -0,0 +1,58 @@
...
+import type { ScreenContextDefinition, ScreenContextActions } from "../state";
+import { screenApiAtom, screenSocketAtom, screenDataAtom } from "../state";
+import { useEvaluate } from "./useEvaluate";
+
+export const useScreenData = (): { apis?: EnsembleAPIModel[] } & {
+  sockets?: EnsembleSocketModel[];
+} & Pick<ScreenContextDefinition, "data"> &
+  Pick<ScreenContextActions, "setData"> & {
+    mockResponses: { [key: string]: EnsembleMockResponse | string | undefined };
+  } => {
```
evshi, line 20: Sorry this isn't your mess but can you help clean up the typing on this and create a new type to capture the return value?

packages/framework/src/hooks/useScreenData.ts
```diff
@@ -0,0 +1,58 @@
...
+      ? apis.reduce(
+          (
+            acc: { [key: string]: EnsembleMockResponse | string | undefined },
+            api,
+          ) => {
+            acc[api.name] = api.mockResponse;
+            return acc;
+          },
+          {},
+        )
```
evshi, line 36: You can use optional chaining here apis?.reduce()

packages/runtime/src/runtime/mock.tsx
```diff
@@ -0,0 +1,38 @@
+import type { EnsembleMockResponse } from "@ensembleui/react-framework";
+import { isObject, isNumber, isEmpty } from "lodash-es";
+
+export const mock = (evaluatedMockResponse: {
```
evshi, line 4: Can we move this into the framework project alongside the rest of the mocking logic?

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -320,6 +341,37 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
...
+        return;
```
evshi, line 371: Rather than repeating this logic can we just move the `mock` into the datafetcher?

evshi (approved):

PR #766, opened 2024-08-05
fix TextInput's regex validator not validating on no input

anserwaseem:
## Describe your changes

**Problem:**
regex: `.{1,}` does not work for "atleast 1 character" when there's no input entered by user - and form is submitted.

**Solution:**
when there's no input entered by user, treat `undefined` value of TextInput as empty string so that regex can be validated successfully.

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Form/TextInput.tsx
+++ b/packages/runtime/src/widgets/Form/TextInput.tsx
@@ -145,9 +145,10 @@ export const TextInput: React.FC<TextInputProps> = (props) => {
 
     if (values?.validator?.regex) {
       rulesArray.push({
-        validator: (_, inputValue: string) => {
+        validator: (_, inputValue?: string) => {
           if (
-            new RegExp(values.validator?.regex || "").test(inputValue || "")
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            new RegExp(values.validator!.regex || "").test(inputValue || "")
           ) {
             return;
           }
--- a/packages/runtime/src/widgets/Form/TextInput.tsx
+++ b/packages/runtime/src/widgets/Form/TextInput.tsx
@@ -143,13 +143,11 @@ export const TextInput: React.FC<TextInputProps> = (props) => {
       });
     }
 
-    if (values?.validator?.regex) {
+    const regex = values?.validator?.regex;
+    if (regex) {
       rulesArray.push({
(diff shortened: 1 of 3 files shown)


evshi:
Why not just use minLength here?

packages/runtime/src/widgets/Form/TextInput.tsx
```diff
@@ -145,8 +145,18 @@ export const TextInput: React.FC<TextInputProps> = (props) => {
 
     if (values?.validator?.regex) {
       rulesArray.push({
-        pattern: new RegExp(values.validator.regex),
-        message: values.validator.regexError || "The field has invalid value",
+        validator: (_, inputValue: string) => {
```
evshi, line 148: Add optional type here.

packages/runtime/src/widgets/Form/TextInput.tsx
```diff
@@ -145,8 +145,18 @@ export const TextInput: React.FC<TextInputProps> = (props) => {
...
+          if (
+            new RegExp(values.validator?.regex || "").test(inputValue || "")
```
evshi, line 150: regex should not be optional at this point because of the if guard.

anserwaseem:
> Why not just use minLength here?

minLength will work fine but won't display the custom `regexError`

packages/runtime/src/widgets/Form/TextInput.tsx
```diff
@@ -145,8 +145,19 @@ export const TextInput: React.FC<TextInputProps> = (props) => {
 
     if (values?.validator?.regex) {
       rulesArray.push({
-        pattern: new RegExp(values.validator.regex),
-        message: values.validator.regexError || "The field has invalid value",
+        validator: (_, inputValue?: string) => {
+          if (
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
evshi, line 150: You shouldn't need this. Typescript should know the type is not undefined/null if you check for it in the if guard

anserwaseem: it still complains
image

evshi: Assign it to a variable and you should be good, e.g.

```ts
    const regex = values?.validator?.regex;
    if (regex) {
```

evshi (approved):

PR #768, opened 2024-08-05
Fix double navigate modal screen

evshi:

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/modal/index.tsx
+++ b/packages/runtime/src/runtime/modal/index.tsx
@@ -118,9 +118,7 @@ export const ModalWrapper: React.FC<PropsWithChildren> = ({ children }) => {
     if (!nextModal || evaluatedOptions.key !== nextModal.key) {
       return;
     }
-    const prevModalStateIndex = modalState.findIndex(
-      (modal) => modal.key === nextModal.key,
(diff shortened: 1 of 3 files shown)


packages/runtime/src/runtime/modal/index.tsx
```diff
@@ -142,23 +163,24 @@ export const ModalWrapper: React.FC<PropsWithChildren> = ({ children }) => {
...
             return oldModalState;
           }
-          return [
+          const nextState = [
             ...oldModalState.slice(0, -1),
             {
               ...oldModalState[oldModalState.length - 1],
               visible: false,
             },
           ];
```
evshi, line 172: @anserwaseem what was the purpose of this?

anserwaseem: it's there such that when a new modal is opened via `navigateModalScreen`, the previous one is hidden

anserwaseem:
what is the exact issue here? is modal created twice in DOM tree?

evshi:
Yes, this triggers two renders of the modal wrapper and causes the modal to be added twice.

anserwaseem (approved):

PR #785, opened 2024-08-13
775 require drawer widget for mobile responsive websites

### LINKED ISSUE #775: Require Drawer Widget for Mobile Responsive Websites
- Following is the dashboard in which I am using this sidebar

image

- This is the mobile responsive design in which I am using Hamburger Button Icon shows the PopupMenu to list the Navbar Items
image

Requirements:
- I need a drawer for the mobile responsive view.
- Drawer should behave in a way that when I click on Hamburger Icon, the Drawer should pop out from left to right as a sidebar but it should overlay the screen.

Example:
image

jkorn8:
## Describe your changes
Added the Drawer widget with the following functionality:
- Customizable height and width
- Customizable positioning of the Drawer on the left, right, top, or bottom
- An onClose event that executes an EnsembleAction
- An id field, where you can open the drawer from any code block in the app using the <id>.openDrawer() or <id>.closeDrawer() functions

[#775](https://github.com/EnsembleUI/ensemble-react/issues/775)

Closes #775-require-drawer-widget-for-mobile-responsive-websites

- [✅] I have performed a self-review of my code
- [✅] I have added tests
- [✅] I have added a changeset `pnpm changeset add`
- [✅] I have added example usage in the kitchen sink app


evshi:
@jkorn8 `Drawer` is not a standalone widget like the others it's an implementation of ViewGroup so it's like the sidebar menu https://docs.ensembleui.com/screens-and-widgets/screen-structure#drawer

packages/framework/src/shared/models.ts
```diff
@@ -42,7 +43,7 @@ export interface EnsembleAppModel {
 
 export interface EnsembleMenuModel {
   id?: string;
-  type: string;
+  type: string; // Type is either 'SideBar' or 'Drawer'
```
evshi, line 46: You can enforce this via string union types, e.g. `'Sidebar' | 'Drawer'`

packages/runtime/src/runtime/screen.tsx
```diff
@@ -33,13 +34,17 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
   const params = new URLSearchParams(search); // query params
   const queryParams: { [key: string]: unknown } = Object.fromEntries(params);
 
-  const mergedInputs = merge(
+  const outletContext = useOutletContext<{ openDrawer: () => void }>();
```
evshi, line 37: Can we generalize this so it's not specific to drawer?

packages/framework/src/parser.ts
```diff
@@ -266,7 +269,7 @@ export const EnsembleParser = {
       | undefined;
     return {
       id: get(menu, [menuType, "id"]) as string | undefined,
-      type: String(menuType),
+      type: menuType.toString(),
```
evshi, line 272: You need to either add optional chaining `?` to prevent an exception. Using the string constructor was for this purpose also.

packages/framework/src/parser.ts
```diff
@@ -462,3 +469,24 @@ const unwrapLanguage = (language: LanguageDTO) => {
     resources: parse(language.content) as { [key: string]: unknown },
   };
 };
+
+const unwrapMenu = (
```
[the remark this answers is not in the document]
evshi: @jkorn8 this is still very specialized for the drawer case.

packages/runtime/src/runtime/entry.tsx
```diff
@@ -51,13 +54,32 @@ export const EnsembleEntry: React.FC<EnsembleEntryProps> = ({
...
+            setDrawerOpen={setDrawerOpen}
+            styles={entry.styles}
+          />
+        ) : null}
+        {entry.type !== "Drawer" && entry.type !== "SideBar" ? (
+          <Alert
+            message={`Menu Type ${entry.type} not supported`}
+            type="error"
+          />
+        ) : null}
```
evshi, line 82: This should be encapsulated in your `EnsembleMenu` component

evshi (changes requested):
It's exciting to see the functionality come together but we need to make some key abstractions to keep the code from getting too specialized.

1. At the parser level, the parser should parse any menu definition into an `EnsembleMenuModel`. It should not know the specifics of any type of menu, i.e. should not reference `Sidebar` or `Drawer`
2. In the view/component layer, the differences in the menu views and handling of errors should be encapsulated by the `EnsembleMenu` component.
3. If it's absolutely necessary to expose the menu functions to the screen, then we need to pass a setter from the `Entry` into the `Menu` so the menu can register all of its function callbacks in a generic way rather than being drawer specific

packages/framework/src/parser.ts
```diff
@@ -462,3 +465,39 @@ const unwrapLanguage = (language: LanguageDTO) => {
...
```
evshi, line 469: What is this the purpose of this function vs `parseMenu`? We should just re-use parseMenu and return a EnsembleMenuModel.

packages/framework/src/shared/models.ts
```diff
@@ -55,6 +56,17 @@ export interface EnsembleMenuModel {
...
+  type: "Drawer"; // More menu types can be added here and should be supported in the future
+  id?: string;
+  styles?: { [key: string]: unknown };
+  onCollapse?: EnsembleAction;
+  title?: string;
+  children?: EnsembleWidget[];
+  header?: EnsembleWidget;
+  footer?: EnsembleWidget;
+}
+
```
evshi, line 69: Get rid of this type and just EnsembleMenuModel.

packages/runtime/src/runtime/ensembleMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+  children?: EnsembleWidget[];
+  header?: EnsembleWidget;
+  footer?: EnsembleWidget;
+  styles?: { [key: string]: unknown };
+}
+
+type Position = "left" | "right" | "top" | "bottom";
+const positions = ["left", "right", "top", "bottom"] as const;
+
+export const EnsembleMenu: React.FC<MenuModel> = (props) => {
```
evshi, line 39: There should only be one component responsible for rendering the menu(s), either this one or `menu.tsx`.

packages/runtime/src/runtime/menu.tsx
```diff
@@ -50,25 +63,47 @@ interface MenuItem {
...
+  searchBoxColor?: TypeColors;
+  iconWidth?: string;
+  iconHeight?: string;
+  width?: string;
+  height?: string;
+  position?: "left" | "right" | "top" | "bottom";
+  onSelectStyles?: {
+    backgroundColor?: TypeColors;
+    borderRadius?: string;
+  };
```
evshi, line 106: Can we share a lot of the common styles here in a shared type?

evshi (commented):
Still a bit further to go. There's a lot of similarity in the menu model up until the view. The only place the implementations should diverge is in the component. There should be one entry point to render a menu and that should then delegate to the specific menu type component.

packages/runtime/src/runtime/menu.tsx
```diff
@@ -160,8 +268,112 @@ export const SideBarMenu: React.FC<MenuBaseProps> = ({ id, ...props }) => {
...
+  const handleClose = (): void => {
+    setOpen(false);
+    onCollapseCallback();
+  };
+
+  const onCollapseAction = useEnsembleAction(onCollapse);
+  const onCollapseCallback = useCallback(() => {
+    if (!onCollapseAction) return;
+    return onCollapseAction.callback();
+  }, [onCollapseAction]);
```
evshi, line 339: There's a lot of shared logic/hooks here that should be hoisted into the parent menu component, and the Drawer and Sidebar should just be responsible for rendering the views. So basically take all the logic in sidebar menu and put it into the parent menu component.

packages/runtime/src/runtime/menu.tsx
```diff
@@ -130,10 +168,80 @@ const CustomLink: React.FC<PropsWithChildren & { item: MenuItem }> = ({
...
+              ? {
+                  openDrawerMenu: () => setCollapsed(true),
+                  closeDrawerMenu: () => setCollapsed(false),
+                }
+              : {
+                  isMenuCollapsed: () => isCollapsed,
+                  setMenuCollapsed: (value: boolean) => setCollapsed(value),
+                }
+          }
+        />
```
evshi, line 209: We can avoid some code dupe here by making the outlet conditional by some property flag.

packages/framework/src/shared/models.ts
```diff
@@ -55,6 +56,8 @@ export interface EnsembleMenuModel {
   styles: { [key: string]: unknown };
 }
 
+export type EnsembleMenuModelType = "SideBar" | "Drawer";
```
evshi, line 59: nit: this could be an enum.

packages/runtime/src/runtime/menu.tsx
```diff
@@ -30,7 +34,20 @@ type TypeColors =
   | "lime"
   | "orange";
 
-interface MenuItem {
+export interface DrawerRef {
```
evshi, line 37: Unused?

packages/runtime/src/runtime/menu.tsx
```diff
@@ -30,7 +34,20 @@ type TypeColors =
...
+  openDrawer: () => void;
+  closeDrawer: () => void;
+  isDrawerOpen: () => boolean;
+}
+
+export interface EnsembleMenuContext {
+  openDrawerMenu?: () => void;
+  closeDrawerMenu?: () => void;
```
evshi, line 45: Can we remove these and just use the menuCollapsed functions?

packages/runtime/src/runtime/menu.tsx
```diff
@@ -130,38 +159,239 @@ const CustomLink: React.FC<PropsWithChildren & { item: MenuItem }> = ({
   );
 };
 
-export const SideBarMenu: React.FC<MenuBaseProps> = ({ id, ...props }) => {
-  const [isCollapsed, setIsCollapsed] = useState<boolean>(false);
-  const { values } = useRegisterBindings({ ...props, isCollapsed }, id, {
-    setCollapsed: setIsCollapsed,
-  });
+export const RenderMenu: React.FC<{
```
evshi, line 162: Can we rename to `EnsembleMenu` and combine with `Menu`?

packages/runtime/src/runtime/menu.tsx
```diff
@@ -130,38 +159,239 @@ const CustomLink: React.FC<PropsWithChildren & { item: MenuItem }> = ({
...
+          flexDirection: "column",
+        }}
+      >
+        <Outlet context={outletContext} />
+      </div>
+    </div>
+  );
+};
+
+export const RenderMenuInView: React.FC<{
```
evshi, line 198: Unused?

packages/runtime/src/runtime/menu.tsx
```diff
@@ -130,38 +159,239 @@ const CustomLink: React.FC<PropsWithChildren & { item: MenuItem }> = ({
...
+        styles={values?.styles || {}}
+      />
+      {values?.footer ? EnsembleRuntime.render([values.footer]) : null}
+    </AntDrawer>
+  );
+};
+
+const ItemsMenu: React.FC<{
+  items: MenuItemProps[];
+  styles: DrawerMenuStyles;
```
evshi, line 382: Update typing to be reflective of both style types.

packages/runtime/src/runtime/menu.tsx
```diff
@@ -130,38 +159,239 @@ const CustomLink: React.FC<PropsWithChildren & { item: MenuItem }> = ({
...
+        selectedItem={selectedItem}
+        setSelectedItem={setSelectedItem}
+        styles={values?.styles || {}}
+      />
+      {values?.footer ? EnsembleRuntime.render([values.footer]) : null}
+    </AntDrawer>
+  );
+};
+
+const ItemsMenu: React.FC<{
```
evshi, line 380: This is a good call breaking this out into its separate component. Can we rename to `MenuItems`?

evshi (commented):
Huge improvements! I see the menu file is getting a bit large. This isn't necessary but you could explore breaking it up into multiples and exporting as a module with index.ts

apps/kitchen-sink/src/ensemble/screens/forms.yaml
```diff
@@ -18,6 +18,42 @@ View:
           onResponse:
             executeCode: |
               console.log('getData', response.body);
+  menu:
```
evshi, line 21: I think this is a super good example to have but can we move it into the layouts page instead?

evshi (approved):

PR #786, opened 2024-08-15
Added support for device object

sagardspeed2:
## Describe your changes
Added support for device object

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/hooks/useScreenContext.tsx
+++ b/packages/framework/src/hooks/useScreenContext.tsx
@@ -1,6 +1,6 @@
 import { Provider, useAtom, useAtomValue, useSetAtom } from "jotai";
-import { useCallback, useContext, useEffect, useMemo } from "react";
-import { clone, debounce, merge } from "lodash-es";
+import { useCallback, useContext } from "react";
+import { clone, merge } from "lodash-es";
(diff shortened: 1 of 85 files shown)


packages/framework/src/hooks/useScreenContext.tsx
```diff
@@ -115,3 +119,27 @@ export const useScreenContext = ():
 
   return { ...screenContext, setData };
 };
+
+const useDeviceObserver = (): void => {
```
evshi, line 123: Can you move this into it's own file?

packages/framework/src/state/device.ts
```diff
@@ -0,0 +1,58 @@
...
+  const setData = useCallback(
+    (values: { [key: string]: unknown }) => {
+      setDataAtom(clone(merge({}, data, values)));
+    },
+    [data, setDataAtom],
+  );
+
+  return {
+    ...data,
+    setData,
```
evshi, line 56: Is there any use for this to be it's own hook? I don't think we allow users to set the device data so can this be combined into `useDeviceObserver`?

evshi:
@sagardspeed2 can you please fix the tests here?

evshi (approved):

PR #796, opened 2024-08-21
modernize PopupMenu widget

### LINKED ISSUE #742: Modernize PopupMenu component
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

anserwaseem:
## Describe your changes
- trigger `onItemSelect` action for both `items` and `item-template`
- add trigger property (by default "click") [possible values: "click" | "hover" | "contextMenu"]
- add `enabled` and `visible` properties in popup menu item, also
- support group menu - see EDL [here](https://github.com/EnsembleUI/ensemble-react/pull/796/files#diff-b34f6a6d35e67887babd0c3fd536a7232c76aa68f211264acee0db1a58057ddeR66)
- support submenus - see EDL [here](https://github.com/EnsembleUI/ensemble-react/pull/796/files#diff-b34f6a6d35e67887babd0c3fd536a7232c76aa68f211264acee0db1a58057ddeR77)
image
- support context menu - see EDL [here](https://github.com/EnsembleUI/ensemble-react/pull/796/files#diff-b34f6a6d35e67887babd0c3fd536a7232c76aa68f211264acee0db1a58057ddeR170)
image

Closes #742 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/PopupMenu.tsx
+++ b/packages/runtime/src/widgets/PopupMenu.tsx
@@ -156,7 +156,7 @@ export const PopupMenu: React.FC<PopupMenuProps> = ({
     ({ key }) => {
       let item;
       if (key.includes("itemTemplate")) {
-        const itemIndex = toNumber(key.at(-1));
+        const itemIndex = toNumber(key.split("_").at(-1));
         item = evaluatedNamedData.namedData[itemIndex];
       } else {
         const itemIndices = key.split("_").filter((_, i) => i > 1);
--- a/packages/runtime/src/widgets/PopupMenu.tsx
+++ b/packages/runtime/src/widgets/PopupMenu.tsx
@@ -1,20 +1,20 @@
-import React, { useEffect, useMemo, useState } from "react";
-import { Menu, Dropdown as AntdDropdown } from "antd";
-import { cloneDeep, get, isEmpty, isObject, isString } from "lodash-es";
+import React, { useCallback, useMemo } from "react";
+import type { MenuProps } from "antd";
+import { Dropdown as AntdDropdown } from "antd";
+import { cloneDeep, isEmpty, isObject, isString, toNumber } from "lodash-es";
 import {
   CustomScopeProvider,
-  defaultScreenContext,
-  evaluate,
   unwrapWidget,
+  useEvaluate,
   useRegisterBindings,
   useTemplateData,
 } from "@ensembleui/react-framework";
 import type {
   CustomScope,
   EnsembleAction,
-  EnsembleWidget,
   Expression,
 } from "@ensembleui/react-framework";
+import type { ItemType } from "antd/es/menu/interface";
 import type {
   EnsembleWidgetProps,
   EnsembleWidgetStyles,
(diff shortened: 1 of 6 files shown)


packages/runtime/src/widgets/PopupMenu.tsx
```diff
@@ -38,99 +49,145 @@ interface PopupMenuStyles {
...
+        item = evaluatedNamedData.namedData[itemIndex];
+      } else {
+        const itemIndices = key.split("_").filter((_, i) => i > 1);
+        if (itemIndices.length > 1) {
+          for (let i = 1; i < itemIndices.length; i += 2) {
+            itemIndices.splice(i, 0, "items");
+          }
+        }
+        item = get(values?.items, itemIndices) as unknown;
+      }
```
evshi, line 169: Does this work if you have double digit items? e.g. 12 items.

anserwaseem: nice catch, ty.

packages/runtime/src/widgets/PopupMenu.tsx
```diff
@@ -29,6 +36,10 @@ const widgetName = "PopupMenu";
 interface PopupMenuItem {
   label: Expression<string> | { [key: string]: unknown };
   value: string;
+  type?: "group";
```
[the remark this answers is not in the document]
anserwaseem: for group menus 
image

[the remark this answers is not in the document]
anserwaseem: ~ it'd be "item" by default

packages/runtime/src/widgets/PopupMenu.tsx
```diff
@@ -25,10 +32,16 @@ import { useEnsembleAction } from "../runtime/hooks/useEnsembleAction";
...
+const DEFAULT_POPUPMENU_TRIGGER = "click";
+const DEFAULT_POPUPMENUITEM_TYPE = "item";
 
 interface PopupMenuItem {
   label: Expression<string> | { [key: string]: unknown };
   value: string;
+  type?: "group" | "item";
+  visible?: boolean;
+  enabled?: boolean;
+  items?: PopupMenuItem[];
```
evshi, line 44: I think we need to be think more about how we can make this API simpler and more re-usable. Can we maybe have nested popup menus instead of requiring different group types?

packages/runtime/src/widgets/PopupMenu.tsx
```diff
@@ -38,99 +51,145 @@ interface PopupMenuStyles {
   borderWidth: string;
 }
 export type PopupMenuProps = {
-  [key: string]: unknown;
   items?: PopupMenuItem[];
   widget?: { [key: string]: unknown };
   onItemSelect?: EnsembleAction;
   showDivider?: boolean | Expression<string>;
+  trigger?: "click" | "hover" | "contextMenu";
+  enabled?: boolean;
```
evshi, line 59: What is the use case for enabled? 

Also I think with item-templates we can deprecate the use of showDivider.

packages/runtime/src/widgets/PopupMenu.tsx
```diff
@@ -38,99 +51,145 @@ interface PopupMenuStyles {
...
+        disabled={values?.enabled === false}
+        menu={{
+          id,
+          items: popupMenuItems,
+          onClick: handleMenuItemClick,
+          style: { overflow: "auto", ...values?.styles },
+        }}
+        trigger={[values?.trigger || DEFAULT_POPUPMENU_TRIGGER]}
+      >
+        <div>{getWidget()}</div>
```
evshi, line 190: What is this extra div element for?

packages/runtime/src/widgets/PopupMenu.tsx
```diff
@@ -38,99 +51,145 @@ interface PopupMenuStyles {
...
+    const widget = cloneDeep(values.widget);
+    const actualWidget = unwrapWidget(widget);
+    return EnsembleRuntime.render([actualWidget]);
+  }, [values?.widget]);
+
+  const handleMenuItemClick = useCallback<NonNullable<MenuProps["onClick"]>>(
+    ({ key }) => {
+      let item;
+      if (key.includes("itemTemplate")) {
+        const itemIndex = toNumber(key.split("_").at(-1));
```
evshi, line 161: This will have issues for double digit items.

packages/runtime/src/widgets/PopupMenu.tsx
```diff
@@ -38,99 +51,145 @@ interface PopupMenuStyles {
...
+        item = evaluatedNamedData.namedData[itemIndex];
+      } else {
+        const itemIndices = key.split("_").filter((_, i) => i > 1);
+        if (itemIndices.length > 1) {
+          for (let i = 1; i < itemIndices.length; i += 2) {
+            itemIndices.splice(i, 0, "items");
+          }
+        }
+        item = get(values?.items, itemIndices) as unknown;
+      }
```
evshi, line 171: Let's memoize a map of the options and their values for quick lookup.

packages/runtime/src/widgets/PopupMenu.tsx
```diff
@@ -38,99 +51,145 @@ interface PopupMenuStyles {
...
-          </React.Fragment>
-        );
-      });
+    const popupItems: MenuProps["items"] = [];
+
+    const items = values?.items;
+    if (items) {
+      const tempItems = items
+        .map((rawItem, index) => getMenuItem(rawItem, index))
+        .filter(Boolean);
```
evshi, line 112: You can use lodash.compact to remove null items.

packages/runtime/src/widgets/PopupMenu.tsx
anserwaseem: we can have both: group-menus and sub-menus (example EDL is already in this PR)

packages/runtime/src/widgets/PopupMenu.tsx
anserwaseem: when `enabled` is set to false; the `widget` provided for PopupMenu will be disabled.
we'd still need showDivider when `items` is being used (and we can't remove items as they have it in [flutter world](https://github.com/EnsembleUI/ensemble/blob/b53a74da0f99b907ed34cc2ea7519c684c77e7d9/modules/ensemble/lib/widget/popup_menu.dart#L33) too)

packages/runtime/src/widgets/PopupMenu.tsx
anserwaseem: how will this be an issue? 
~ already showing 20 items [here](https://github.com/EnsembleUI/ensemble-react/pull/796/files#diff-b34f6a6d35e67887babd0c3fd536a7232c76aa68f211264acee0db1a58057ddeR58)

packages/runtime/src/widgets/PopupMenu.tsx
anserwaseem: can use fragment like other places.
~ was using it as our runtime render method returns an array of ReactNode

packages/runtime/src/widgets/PopupMenu.tsx
evshi: Yes, I meant without requiring special properties for both.

anserwaseem: type: "item" is the default value through which simple menu and submenu can be created.
Just for group menu, we'll have to pass in type: "group" as we're doing in [Dropdown widget](https://github.com/EnsembleUI/ensemble-react/blob/071fe2c8adfbcc3c3cccf544bb7bd76cd247ac06/packages/runtime/src/widgets/Form/Dropdown.tsx#L46)

evshi (approved):

PR #823, opened 2024-09-09
hotfix: screen level style apply support

sagardspeed2:
## Describe your changes
screen level style apply support

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/body.tsx
+++ b/packages/runtime/src/runtime/body.tsx
@@ -1,5 +1,5 @@
 import {
-  useEvaluate,
+  useRegisterBindings,
   type EnsembleFooterModel,
   type EnsembleHeaderModel,
   type EnsembleWidget,
@@ -8,6 +8,7 @@ import { ConfigProvider } from "antd";
 import { WidgetRegistry } from "../registry";
 // eslint-disable-next-line import/no-cycle
 import { Column } from "../widgets/Column";
+import { isUndefined, omitBy } from "lodash-es";
 
 interface EnsembleBodyProps {
   body: EnsembleWidget;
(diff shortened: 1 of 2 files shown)


anserwaseem:
@sagardspeed2 what's the issue that this PR resolves?

sagardspeed2:
@anserwaseem previously fontSize was passing as undefined if not mentioned in themes or manually in styles and due to that, radio buttons and some places of forms was not showing properly, so to fix that it is required to only override font size if it was passed

packages/runtime/src/runtime/body.tsx
```diff
@@ -18,7 +19,7 @@ interface EnsembleBodyProps {
 }
 
 export const EnsembleBody: React.FC<EnsembleBodyProps> = ({ body, styles }) => {
-  const evaluatedStyles = useEvaluate({ styles });
+  const { values } = useRegisterBindings({ styles });
```
anserwaseem, line 22: why move to using `useRegisterBindings` hook?

anserwaseem (changes requested):

packages/runtime/src/runtime/body.tsx
sagardspeed2: because in style if we passed classname, then we need to extract css from that classname, useEvaluate only evaluate it as js and resolve variable values

sagardspeed2: @anserwaseem

sagardspeed2: @anserwaseem  That is not needed, because if we include named and classname handling into useEvaluate, then it will become same expensive as useRegisterBindings,

and that's why I am only passing the styles into useRegisterBinding to avoid unnecessary load

sagardspeed2: it will just fetch the styles from the theme, sometimes there are multiple level of expressions or variables assigned, I was in this situation some months ago, me and Evan spend much time to find a better solution and then we decided this, If you want to test, you can test in your side by placing variables into themes and also in screen yaml

PR #824, opened 2024-09-10
feat: modernaization radio widget

### LINKED ISSUE #817: Modernize Radio widget
Modernize radio widget per [docs](https://docs.ensembleui.com/widgets/radio-group) 

**Additional context**

sagardspeed2:
## Describe your changes
Added item-template support in Radio widget

Closes #817 

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/hooks/useTemplateData.ts
+++ b/packages/framework/src/hooks/useTemplateData.ts
@@ -1,6 +1,6 @@
 import { useMemo } from "react";
 import { type Atom, atom, useAtomValue } from "jotai";
-import { get, isString, map, merge } from "lodash-es";
+import { isString, map, merge } from "lodash-es";
 import { createBindingAtom } from "../evaluate";
 import { isExpression, type Expression } from "../shared/common";
 import { useCustomScope } from "./useCustomScope";
@@ -11,7 +11,6 @@ export interface TemplateDataProps {
   data?: Expression<TemplateData>;
   name?: string;
   context?: unknown;
-  value?: Expression<string>;
 }
 
 /**
@@ -26,7 +25,6 @@ export const useTemplateData = ({
   data,
   name = "data",
   context,
-  value,
 }: TemplateDataProps): {
   rawData: TemplateData;
   namedData: object[];
@@ -56,17 +54,6 @@ export const useTemplateData = ({
 
   return {
     rawData,
-    namedData: map(evaluatedNamedData.namedData, (val) => ({
-      ...val,
-      _ensembleValue: value
-        ? (get(
-            val,
-            value
-              .toString()
-              // eslint-disable-next-line prefer-named-capture-group
-              .replace(/['"]?\$\{([^}]*)\}['"]?/g, "$1"),
-          ) as string | number)
-        : "",
-    })),
+    namedData: evaluatedNamedData.namedData,
   };
 };
--- a/packages/framework/src/hooks/useTemplateData.ts
+++ b/packages/framework/src/hooks/useTemplateData.ts
@@ -1,6 +1,6 @@
 import { useMemo } from "react";
 import { type Atom, atom, useAtomValue } from "jotai";
-import { isString, map, merge } from "lodash-es";
+import { get, isString, map, merge } from "lodash-es";
 import { createBindingAtom } from "../evaluate";
 import { isExpression, type Expression } from "../shared/common";
 import { useCustomScope } from "./useCustomScope";
@@ -11,6 +11,7 @@ export interface TemplateDataProps {
   data?: Expression<TemplateData>;
   name?: string;
   context?: unknown;
+  value?: Expression<string>;
 }
 
 /**
@@ -25,6 +26,7 @@ export const useTemplateData = ({
   data,
   name = "data",
   context,
+  value,
 }: TemplateDataProps): {
   rawData: TemplateData;
   namedData: object[];
@@ -46,13 +48,25 @@ export const useTemplateData = ({
   }, [customScope, data, isDataString, context]);
   const rawData = useAtomValue(dataAtom);
   const namedData = useMemo(
-    () => map(rawData, (value: unknown) => ({ [name]: value })),
+    () => map(rawData, (val: unknown) => ({ [name]: val })),
     [name, rawData],
   );
+
   const evaluatedNamedData = useEvaluate({ namedData });
 
   return {
     rawData,
-    namedData: evaluatedNamedData.namedData,
+    namedData: map(evaluatedNamedData.namedData, (val) => ({
+      ...val,
+      value: value
+        ? (get(
+            val,
+            value
+              .toString()
+              // eslint-disable-next-line prefer-named-capture-group
+              .replace(/['"]?\$\{([^}]*)\}['"]?/g, "$1"),
+          ) as string | number)
+        : "",
+    })),
   };
 };
--- a/packages/framework/src/hooks/useTemplateData.ts
+++ b/packages/framework/src/hooks/useTemplateData.ts
@@ -58,7 +58,7 @@ export const useTemplateData = ({
     rawData,
     namedData: map(evaluatedNamedData.namedData, (val) => ({
       ...val,
-      value: value
+      _ensembleValue: value
         ? (get(
             val,
             value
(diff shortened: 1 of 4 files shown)


packages/runtime/src/widgets/Form/Radio.tsx
```diff
@@ -21,15 +33,42 @@ export type RadioWidgetProps = FormInputProps<string> &
...
+  // update form field initial value
+  const formInstance = Form.useFormInstance();
+
+  useEffect(() => {
+    if (formInstance) {
+      formInstance.setFieldsValue({
+        [values?.id ?? values?.label ?? ""]: value,
+      });
+    }
+  }, [value, formInstance]);
```
evshi, line 66: Why is this needed?

packages/runtime/src/widgets/Form/Radio.tsx
```diff
@@ -38,19 +77,58 @@ export const RadioWidget: React.FC<RadioWidgetProps> = (props) => {
...
+          style: getStyle(item.styles as object),
+        });
       });
     }
-  }, [value, formInstance]);
+
+    if (isObject(itemTemplate) && !isEmpty(namedData)) {
+      namedData.forEach((item: unknown) => {
+        const typedItem = get(item, itemTemplate.name) as CustomScope;
+        const evaluateValue = evaluate<string | number>(
```
evshi, line 109: Do not use.

evshi (changes requested):

packages/runtime/src/widgets/Form/Radio.tsx
sagardspeed2: @evshi this is needed to set the initial values of radio widget in the form instance, so it will validate that values or we can get those values on form submit time.

we did this in all form elements like, dropdown, radio, checkbox, textInput, Date and so on

packages/runtime/src/widgets/Form/Radio.tsx
```diff
@@ -21,15 +31,42 @@ export type RadioWidgetProps = FormInputProps<string> &
...
-    props.id,
+    { ...rest, initialValue: rest.value, value, widgetName },
+    rest.id,
     { setValue },
   );
-  const action = useEnsembleAction(props.onChange);
 
+  // update initial value
+  useEffect(() => {
+    setValue(values?.initialValue?.toString());
```
evshi, line 52: Value may not be a string so don't force it to one.

packages/runtime/src/widgets/Form/Radio.tsx
```diff
@@ -38,19 +75,60 @@ export const RadioWidget: React.FC<RadioWidgetProps> = (props) => {
...
+    if (isObject(itemTemplate) && !isEmpty(namedData)) {
+      namedData.forEach((item: unknown) => {
+        const typedItem = get(item, itemTemplate.name) as CustomScope;
+        const evaluateValue = get(
+          item,
+          itemTemplate.value
+            .toString()
+            // eslint-disable-next-line prefer-named-capture-group
+            .replace(/['"]?\$\{([^}]*)\}['"]?/g, "$1"),
+        ) as string | number;
```
evshi, line 113: Again this is forcing value as a string?

evshi (changes requested):
Please write some tests that include values of different types - string, number, object.

packages/runtime/src/widgets/Form/Radio.tsx
sagardspeed2: @evshi here, `itemTemplate.value` will be always string, for example,

```yaml
item-template:
    data: getData.body.users.slice(0, 4)
    name: user
    value: ${user.email}
    template:
      Row:
        styles:
          gap: 10px
          paddingBlock: 10px
          alignItems: center
        children:
          - Image:
              width: 25
              height: 25
              source: ${user.image}
          - Text:
              text: ${user.firstName} ${user.lastName}
```

where user.email will be always string and indicate that which value need to assign that iterate, here it says that use `user.email` as a value.

evshi: Oh I misread this. You shouldn't need to find the evaluated value - `useItemTemplate` does that for you in the named objects.

evshi: @sagardspeed2 let's extend useTemplateData to handle an optional `value` prop from itemTemplate so it evaluates for you.

packages/framework/src/hooks/useTemplateData.ts
```diff
@@ -46,13 +48,25 @@ export const useTemplateData = ({
...
   );
+
   const evaluatedNamedData = useEvaluate({ namedData });
 
   return {
     rawData,
-    namedData: evaluatedNamedData.namedData,
+    namedData: map(evaluatedNamedData.namedData, (val) => ({
+      ...val,
+      value: value
```
evshi, line 61: Pass this into useEvaluate. Make the key `_ensembleValue` to avoid conflicts with the evaluated keys.

evshi (changes requested):

packages/framework/src/hooks/useTemplateData.ts
sagardspeed2: @evshi that will not work. because of, as per your suggestion,

```js
  const namedData = useMemo(
    () =>
      map(rawData, (val: unknown) => ({ [name]: val, _ensembleValue: value })),
    [name, rawData],
  );

  const evaluatedNamedData = useEvaluate({ namedData });
```

when I will pass `_ensembleValue` in useEvaluate like this way, where yaml is like,

```yaml
    Radio:
      id: radioGroup
      label: Radio Group With Template
      item-template:
        data: getData.body.users.slice(0, 4)
        name: user
        value: ${user.email}
        template:
          Row:
            children:
              - Image:
                  width: 25
                  height: 25
                  source: ${user.image}
              - Text:
                  text: ${user.firstName} ${user.lastName}
                  
```

so each object of namedData will become like this before getting evaluated by `useEvaluate`

```js
{
user: {...},
_ensembleValue: ${user.email}
}
```

then when we pass this to useEvaluate, it will extract expressions like this,
image

and then this expression will convert into bindings and will be try to evaluate with ensemble storage, ensemble env, defaultScreenContext, ensemble user, ensemble secrets, device and so on.

and eventually it return the `_ensembleValue` as `null` because this value is depend on `user` of `namedData` which is not available in context.

sagardspeed2:
@evshi I tried to update `useItemTemplateData` and `useEvalaute` hook to achieve your suggestion idea.
https://github.com/EnsembleUI/ensemble-react/pull/852

evshi:
Please close this PR if #852 covers all the changes.

packages/runtime/src/widgets/Form/Radio.tsx
```diff
@@ -38,19 +98,55 @@ export const RadioWidget: React.FC<RadioWidgetProps> = (props) => {
...
+  // extract template data
+  const { namedData } = useTemplateData({
+    data: itemTemplate?.data,
+    name: itemTemplate?.name,
+  });
 
-  useEffect(() => {
-    setValue(values?.initialValue?.toString());
-  }, [values?.initialValue]);
+  const RadioComponent = useMemo(() => withTemplate(Radio), []);
```
evshi, line 107: Remove memo.

packages/runtime/src/widgets/Form/Radio.tsx
```diff
@@ -38,19 +98,55 @@ export const RadioWidget: React.FC<RadioWidgetProps> = (props) => {
...
+      ...(styles?.visible === false ? { display: "none" } : undefined),
+    });
+
+    if (values?.items) {
+      values.items.forEach((item) => {
+        radioOptions.push({
+          item,
+          children: item.label,
+          disabled: values.enabled === false || item.enabled === false,
+          value: String(item.value),
```
evshi, line 124: Do not coerce value to string.

packages/runtime/src/widgets/Form/Radio.tsx
```diff
@@ -38,19 +98,55 @@ export const RadioWidget: React.FC<RadioWidgetProps> = (props) => {
...
       });
     }
-  }, [value, formInstance]);
+
+    if (isObject(itemTemplate) && !isEmpty(namedData)) {
+      map(namedData, (item: { [key: string]: unknown }) => {
+        const typedItem = get(item, itemTemplate.name) as CustomScope;
+        radioOptions.push({
+          item,
+          disabled: values?.enabled === false || typedItem.enabled === false,
```
evshi, line 135: Don't check `typedItem` for enabled - it's data it can't be guaranteed to have enabled prop.

evshi (changes requested):

evshi (approved):

PR #833, opened 2024-09-16
added range support in slider

### LINKED ISSUE #789: Slider that allows you to select a range
I need to use a slider for one of my screens but there's no option to have more than one value for the slider. For example if I wanted to select a range between 10 and 30, there's no way to do that.

For a solution, I just want a way to be able to have two values!

I could definitely use JavaScript but I think that's doing a lot more than is necessary.

For context, the screen I'm making needs a slider because the user needs to select a distance they're comfortable with, and a price range they're comfortable with.

sagardspeed2:
## Describe your changes
Added range support in slider

Closes #789 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Slider.tsx
+++ b/packages/runtime/src/widgets/Slider.tsx
@@ -18,7 +18,7 @@ interface SliderStyles {
 }
 
 export type SliderProps = {
-  initialValue?: number | number[];
+  initialValue?: number | [number, number];
   onChange?: EnsembleAction;
   onComplete?: EnsembleAction;
   min?: number;
@@ -28,11 +28,11 @@ export type SliderProps = {
   dots?: boolean;
   divisions?: number;
   range?: boolean;
-} & FormInputProps<number | number[]> &
+} & FormInputProps<number | [number, number]> &
   EnsembleWidgetProps<SliderStyles>;
 
 const SliderWidget: React.FC<SliderProps> = (props) => {
-  const [value, setValue] = useState<number | number[]>();
+  const [value, setValue] = useState<number | [number, number]>();
   const { values, rootRef } = useRegisterBindings(
     { ...props, value, widgetName },
     props.id,
@@ -51,7 +51,7 @@ const SliderWidget: React.FC<SliderProps> = (props) => {
   const onCompleteAction = useEnsembleAction(props.onComplete);
 
   const onChangeActionCallback = useCallback(
-    (newValue: number | number[]) => {
+    (newValue: number | [number, number]) => {
       if (!onChangeAction) {
         return;
       }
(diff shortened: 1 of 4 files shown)


packages/runtime/src/widgets/Slider.tsx
```diff
@@ -93,11 +102,13 @@ const SliderWidget: React.FC<SliderProps> = (props) => {
           dots={values?.dots}
           max={values?.max}
           min={values?.min}
-          onAfterChange={handleAfterChangeComplete}
           onChange={handleChange}
+          onChangeComplete={handleAfterChangeComplete}
+          range={values?.range || {}}
```
evshi, line 107: Why is this an object? Just do boolean coercion

packages/runtime/src/widgets/Slider.tsx
```diff
@@ -26,23 +27,31 @@ export type SliderProps = {
...
   dots?: boolean;
   divisions?: number;
+  range?: boolean;
 } & FormInputProps<number | number[]> &
   EnsembleWidgetProps<SliderStyles>;
 
 const SliderWidget: React.FC<SliderProps> = (props) => {
-  const [value, setValue] = useState(props.value);
-  const { values } = useRegisterBindings(
+  const [value, setValue] = useState<number | number[]>();
```
evshi, line 35: all values and initial value should have type `number | [number, number]`. You can alias this for easier reference but it needs to be consistent.

evshi (changes requested):

packages/runtime/src/widgets/Slider.tsx
sagardspeed2: @evshi  The `Range` props in the `Slider` allows for more than two nodes, meaning we need to define its value as a number[]. According to the RCSlider documentation and types, the value must be passed as a number[], enabling the creation of multiple nodes within the same slider.

image

packages/runtime/src/widgets/Slider.tsx
```diff
@@ -83,23 +92,40 @@ const SliderWidget: React.FC<SliderProps> = (props) => {
     }
   `;
 
+  const sliderProps = useMemo(
```
evshi, line 95: What problem is this solving?

packages/runtime/src/widgets/Slider.tsx
evshi: Can you provide a link to the documentation? We need to decouple our component interface from the underlying implementation otherwise this will create problems if we need to change the implementation.

sagardspeed2: @evshi Here is the documentation link of `Slider` widget
https://ant.design/components/slider

sagardspeed2:
@evshi  Here is the documentation link of Slider widget
https://ant.design/components/slider

packages/runtime/src/widgets/Slider.tsx
evshi: value clearly says `number | [number, number]` - let's focus on the simple case first.

packages/runtime/src/widgets/Slider.tsx
evshi: This useMemo isn't doing anything let's just use a raw object.

evshi (approved):

PR #852, opened 2024-09-30
fix: added _ensembleValue in itemTemplateData in evaluate stage

sagardspeed2:

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/hooks/useNavigateModal.tsx
+++ b/packages/runtime/src/runtime/hooks/useNavigateModal.tsx
@@ -56,7 +56,7 @@ export const useNavigateModalScreen: EnsembleActionHook<
       const inputs =
         !isString(action) && action.inputs ? cloneDeep(action.inputs) : {};
       if (screenContext) {
-        const expressionMap: string[][] = [];
+        const expressionMap: [string, string, object][] = [];
         findExpressions(inputs, [], expressionMap);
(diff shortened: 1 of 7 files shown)


packages/runtime/src/runtime/hooks/useNavigateModal.tsx
```diff
@@ -56,7 +56,7 @@ export const useNavigateModalScreen: EnsembleActionHook<
       const inputs =
         !isString(action) && action.inputs ? cloneDeep(action.inputs) : {};
       if (screenContext) {
-        const expressionMap: string[][] = [];
+        const expressionMap: [string, string, object][] = [];
```
evshi, line 59: Why is this needed?

evshi (changes requested):

packages/runtime/src/runtime/hooks/useNavigateModal.tsx
sagardspeed2: @evshi previously is was the array of string only, and I need to push the object in that array whenever it find the expression

evshi: Why do you need the object...?

sagardspeed2: You mean, the types is wrong or I can't use object whenever it found the expression ?

PR #862, opened 2024-10-08
fix: added useEffect cleanup in useRegisterBindings

### LINKED ISSUE #861: Cannot call button setLoading multiple times
**Describe the bug**
Put a button in a showDialog with an id and call setLoading. Works first time but not the second time. The issue is the setLoading invokable function becomes stale.

We did a short term fix for forceState but this is no longer needed . We just need to compare function equality.

**To Reproduce**

1. Create a dialog with a button with an ID
2. Call setLoading on the button
3. Close modal
4. Open modal
5. Call setLoading on the button again
6. Button does not enter loading state

**Expected behavior**
Button should enter loading state no matter how many times modal renderes

**Screenshots**

**Additional context**

sagardspeed2:
## Describe your changes
Issue related widget states become stale on component re-render

for example,
added a button on showDialog, opened showDialog and set that button loader to true, close the showDialog and open again, try again to set loader, and this time nothing happen due to setter function become stale.

### Screenshots [Optional]
after:

https://github.com/user-attachments/assets/044425a9-82c9-4c46-a669-b7ea5d25d922

Closes #861 

- [ ] I have added tests


packages/framework/src/hooks/useRegisterBindings.ts
```diff
@@ -88,6 +88,10 @@ export const useRegisterBindings = <T extends { [key: string]: unknown }>(
     }
   }, [options?.forceState]);
 
+  useEffect(() => {
+    setWidgetState({});
+  }, [setWidgetState]);
```
sagardspeed2, line 93: @evshi this basically reset the values of that widget on widget unmounted. and thus on next render of that element will re-render and again that in `widgetState`

evshi:
@sagardspeed2 this will create infinite renders and cause other dependent expressions to fail. See the failing test. Just update the invokables each time.

packages/framework/src/hooks/useRegisterBindings.ts
```diff
@@ -88,6 +88,18 @@ export const useRegisterBindings = <T extends { [key: string]: unknown }>(
...
+    return () => {
+      setWidgetState({
+        values: newValues,
+        invokable: {
+          id: resolvedWidgetId,
+          methods: { ...methods, getId: () => resolvedWidgetId },
+        },
+      });
+    };
+  }, [setWidgetState]);
```
evshi, line 101: This doesn't seem right. This pushes a state update when the component unmounts, which triggers more renders on dependent widgets. We should limit use of setWidgetState.

Update the comparator on line 61 to better handle functions.

packages/runtime/src/widgets/ImageCropper.tsx
```diff
@@ -38,15 +38,12 @@ export const ImageCropper: React.FC<ImageCropperProps> = (props) => {
...
-      }
-      onCroppedAction.callback({ file });
-    },
-    [onCroppedAction],
-  );
+  const onCroppedActionCallback = useCallback((file: unknown) => {
+    if (!onCroppedAction) {
+      return;
+    }
+    onCroppedAction.callback({ file });
```
evshi, line 45: So this is an example where you do need the action in the dependency, or at the last the callback. If the callback hook changes any of the dependencies this does need to get recomputed.

packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -215,7 +215,7 @@ export const DataGrid: React.FC<GridProps> = (props) => {
     if (onScrollEndAction) {
       onScrollEndAction.callback();
     }
-  }, [onScrollEndAction]);
+  }, []);
```
evshi, line 218: This needs to be at least onScrollEndAction.callback, same for below.

packages/runtime/src/widgets/Button.tsx
```diff
@@ -33,16 +33,13 @@ export type ButtonProps = {
...
-    },
-    [action],
-  );
+  const onClickCallback = useCallback((e?: MouseEvent) => {
+    e?.stopPropagation();
+    if (!action) {
+      return;
+    }
+    action.callback();
+  }, []);
```
evshi, line 42: This should be action.callback in the dependency array.

packages/framework/src/hooks/useRegisterBindings.ts
```diff
@@ -49,16 +56,23 @@ export const useRegisterBindings = <T extends { [key: string]: unknown }>(
   );
 
   useEffect(() => {
-    // Improves performance greatly: o need to store state in global if there's no explicit ID to reference it with
```
evshi, line 52: Keep this comment.

packages/framework/src/hooks/useRegisterBindings.ts
```diff
@@ -15,6 +15,13 @@ export interface RegisterBindingsResult<T> {
   rootRef: RefCallback<never>;
 }
 
+// eslint-disable-next-line @typescript-eslint/ban-types
+const areFunctionsEqual = (func1?: Function, func2?: Function): boolean => {
+  if (!func1 && !func2) return true;
+  if (!func1 || !func2) return false;
```
evshi, line 21: Why are these even needed? You can just use triple equal?

packages/framework/src/hooks/useRegisterBindings.ts
```diff
@@ -49,16 +56,23 @@ export const useRegisterBindings = <T extends { [key: string]: unknown }>(
...
+      // Check if the method references are equal or if their string representations are the same
+      return areFunctionsEqual(fromMethods, fromWidgets);
+    });
+
     if (
       (options?.comparator
         ? options.comparator(newValues, widgetState?.values)
         : isEqual(newValues, widgetState?.values)) &&
-      isEqual(keys(methods), keys(widgetState?.invokable?.methods))
+      prevStateCheck
```
evshi, line 75: Combine this function check with the values check.

packages/framework/src/hooks/useRegisterBindings.ts
```diff
@@ -70,7 +84,6 @@ export const useRegisterBindings = <T extends { [key: string]: unknown }>(
   }, [
     methods,
     resolvedWidgetId,
-    setWidgetState,
```
evshi, line 73: Why remove this?

packages/runtime/src/widgets/__tests__/Button.test.tsx
```diff
@@ -228,4 +230,198 @@ test("test button loading using bindings", async () => {
...
+            id="test-button"
+            // eslint-disable-next-line no-template-curly-in-string
+            label="${ensemble.storage.get('btnLabel')}"
+            options={{ debounceMs: 100 }}
+          />
+          <Button
+            label="Change button label"
+            onTap={{
+              executeCode:
+                "ensemble.storage.set('btnLabel', 'Ensemble Button')",
```
evshi, line 404: You need a test that checks the render count when one of the bindings on the actions change.

packages/runtime/src/widgets/__tests__/Button.test.tsx
```diff
@@ -228,4 +230,198 @@ test("test button loading using bindings", async () => {
...
+      );
+    };
+
+    const { getByTestId } = render(<Parent />);
+
+    // Initial render
+    expect(renderCount.current).toBe(1);
+
+    // Update parent state
+    // eslint-disable-next-line @typescript-eslint/require-await
```
evshi, line 279: These shouldn't be needed.

packages/framework/src/hooks/__tests__/useRegisterBindings.test.ts
```diff
@@ -322,3 +338,66 @@ test("evaluates flutter style hex codes expressions", () => {
...
+  // Rerender with the same dependency
+  rerender({ dependency: 1 });
+
+  const secondCallback = result.current;
+
+  // Check that the callback reference remains the same
+  expect(firstCallback.toString()).toBe(secondCallback.toString());
+});
+
+test("should keep the same reference for all functions inside an object when dependencies do not change", () => {
```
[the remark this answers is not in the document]
evshi: @sagardspeed2 please add this test.

packages/framework/src/hooks/useRegisterBindings.ts
sagardspeed2: Not needed and not used in that useEffect

packages/runtime/src/runtime/hooks/useCloseAllModalScreens.ts
```diff
@@ -8,7 +8,9 @@ export const useCloseAllScreens: EnsembleActionHook<
 > = () => {
   const { closeAllScreens } = useContext(ModalContext) || {};
 
-  const closeAllModalScreens: () => void = () => closeAllScreens?.();
+  const callback = useCallback(() => {
+    closeAllScreens?.();
+  }, [closeAllScreens]);
 
-  return { callback: closeAllModalScreens };
+  return useMemo(() => ({ callback }), [callback]);
```
evshi, line 15: This isn't doing anything

evshi, line 15: This isn't doing anything

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -146,7 +146,7 @@ export const useExecuteCode: EnsembleActionHook<
     { modalContext, render: EnsembleRuntime.render, EnsembleScreen },
   );
 
-  return { callback: execute };
+  return useMemo(() => ({ callback: execute }), [execute]);
```
evshi, line 149: Not doing anything.

evshi (changes requested):
Remove all the useMemos where the command is also the dependency that doesn't do anything.

packages/framework/src/hooks/useRegisterBindings.ts
```diff
@@ -15,6 +15,35 @@ export interface RegisterBindingsResult<T> {
   rootRef: RefCallback<never>;
 }
 
+// eslint-disable-next-line @typescript-eslint/ban-types
+const areFunctionsEqual = (func1?: Function, func2?: Function): boolean => {
+  return func1 === func2;
+};
```
evshi, line 21: Can you just inline this so we don't need the linter comment?

packages/runtime/src/runtime/hooks/useCloseAllModalScreens.ts
```diff
@@ -8,7 +8,9 @@ export const useCloseAllScreens: EnsembleActionHook<
 > = () => {
   const { closeAllScreens } = useContext(ModalContext) || {};
 
-  const closeAllModalScreens: () => void = () => closeAllScreens?.();
+  const closeAllModalScreens = useCallback(() => {
+    closeAllScreens?.();
+  }, []);
```
evshi, line 13: This should memoize the return object like we discussed.

packages/framework/src/hooks/useScreenData.ts
```diff
@@ -49,11 +51,14 @@ export const useScreenData = (): {
...
   );
 
-  return {
-    apis,
-    sockets,
-    data,
-    setData,
-    mockResponses,
-  };
+  return useMemo(
```
evshi, line 54: Why is this needed again?

evshi: I think for this hook, unlike the action hooks, since it is used so rarely that we don't need to memoize the return object but fix the usages instead.

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -311,9 +311,9 @@ export const useConnectSocket: EnsembleActionHook<ConnectSocketAction> = (
   }, [
     screenData,
     socket,
-    onSocketConnectAction,
-    onMessageReceiveAction,
-    onSocketDisconnectAction,
+    onSocketConnectAction?.callback,
+    onMessageReceiveAction?.callback,
+    onSocketDisconnectAction?.callback,
```
evshi, line 316: These changes are fine but preferably we need to memoize the action return object.

packages/framework/src/hooks/useEvaluate.ts
```diff
@@ -83,5 +83,7 @@ export const useEvaluate = <T extends Record<string, unknown>>(
     return result;
   }, [updatedValues]);
 
-  return merge({}, updatedValues, hexCodes);
+  return useMemo(() => {
+    return merge({}, updatedValues, hexCodes);
+  }, [updatedValues, hexCodes]);
```
evshi, line 88: What is the purpose of this?

packages/framework/src/hooks/__tests__/useRegisterBindings.test.ts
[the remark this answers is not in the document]
evshi: @sagardspeed2 you need to add this test.

packages/runtime/src/widgets/__tests__/Button.test.tsx
evshi: @sagardspeed2 please clean this up

packages/runtime/src/widgets/__tests__/Button.test.tsx
evshi: @sagardspeed2 did you add this test somewhere?

evshi (commented):
Almost there just a few tests to write to confirm results and clean up

packages/framework/src/hooks/__tests__/useRegisterBindings.test.ts
```diff
@@ -322,3 +341,194 @@ test("evaluates flutter style hex codes expressions", () => {
...
+  );
+
+  // Rerender with new methods
+  rerender({
+    values: testValues,
+    methods: { updateCount: fn2 },
+  });
+
+  await waitFor(() => {
+    const finalScreen = store.get(screenAtom);
```
evshi, line 526: Can you assert the resulting methods that get returned from `useRegisterBindings` is also updated?

sagardspeed2: @evshi we are not getting resulting methods returned from `useRegisterBindings`, we only get `id`, `rootRef` and `values` from `useRegisterBindings`

evshi (approved):

PR #874, opened 2024-10-18
fix: button loader binding evaluation

### LINKED ISSUE #873: Button loading binding does not work
**Describe the bug**
Using an expression for the button loading prop does not update properly. This is because the prop is never registered as a binding it is overwritten with local state.

**To Reproduce**

1. Create a button with a loading prop and bind to some expression
2. Trigger expression to true
3. Button loading state does not update

**Expected behavior**
Button loading state should update

**Screenshots**

**Additional context**
**Write tests!!!**

sagardspeed2:
## Describe your changes
Button loader binding evaluation

Closes #873 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/__tests__/Button.test.tsx
+++ b/packages/runtime/src/widgets/__tests__/Button.test.tsx
@@ -1,0 +1,118 @@
+/* eslint-disable react/no-children-prop */
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import { BrowserRouter } from "react-router-dom";
+import { Button } from "../Button";
+import { createCustomWidget } from "../../runtime/customWidget";
+import { Column } from "../Column";
+
+describe("Button", () => {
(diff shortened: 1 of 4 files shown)


packages/runtime/src/widgets/__tests__/Button.test.tsx
```diff
@@ -0,0 +1,118 @@
+/* eslint-disable react/no-children-prop */
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import { BrowserRouter } from "react-router-dom";
+import { Button } from "../Button";
+import { createCustomWidget } from "../../runtime/customWidget";
+import { Column } from "../Column";
+
+describe("Button", () => {
+  test("test button loading using setLoading", async () => {
```
evshi, line 10: You need to add a test where `loading` is initialized with a binding but then changed via `setLoading`

evshi (approved):

PR #880, opened 2024-10-22
fix: multiselect select/deselect issue

### LINKED ISSUE #879: Not able to select/deselect options in MultiSelect
**Describe the bug**
Please refer to the following video

https://github.com/user-attachments/assets/0e713dab-0a03-4754-b69a-40f521224783

**Introduced in Version**
@ensembleui/react-runtime: 0.2.9

sagardspeed2:
## Describe your changes
MultiSelect select / deselect issue fix

Closes #879 

- [x] I have added tests


packages/runtime/src/widgets/Form/__tests__/MultiSelect.test.tsx
```diff
@@ -196,19 +205,45 @@ describe("MultiSelect Widget", () => {
...
 
     userEvent.click(screen.getByRole("combobox"));
-    fireEvent.click(screen.getByTitle("Option 2"));
-    fireEvent.click(screen.getByTitle("Option 4"));
+    userEvent.click(screen.getByTitle("Option 2"));
+    userEvent.click(screen.getByTitle("Option 4"));
     userEvent.click(screen.getByRole("combobox"));
 
+    // Wait for the combobox to reflect the selected values
+    await waitFor(() => {
```
evshi, line 213: Why was this test passing before, and how does this change prevent this issue from occurring?

sagardspeed2: Yes, it found multiple elements with same text, one in input area where selected values will show and one from dropdown list, I confirmed this with the element selector and both are different element with same text. I also tried to close the dropdown list after selecting the option with fireEvent and userEvent. the dropdown list is closing but it is still part of document

packages/runtime/src/widgets/Form/MultiSelect.tsx
```diff
@@ -67,9 +73,17 @@ const MultiSelect: React.FC<MultiSelectProps> = (props) => {
...
+  // check and load initial values
   useEffect(() => {
-    if (isArray(values?.initialValue)) {
+    // compare previous initial value with current render initial value
+    if (
+      !isEqual(prevInitialValue.current, values?.initialValue) &&
+      isArray(values?.initialValue)
+    ) {
+      prevInitialValue.current = values?.initialValue || [];
       setSelectedValues(values?.initialValue);
```
evshi, line 87: Comments should explain why this code is here, not what it does. The code already says what it does.

This should say, `We need to deep compare the value of the binding to tell when it has changed because it's an array`

evshi (approved):

PR #883, opened 2024-10-24
fix: mask property not working

### LINKED ISSUE #878: TextInput: mask property not working if user copy pastes unmasked number
Ref: https://atlashealth.atlassian.net/browse/TRINITY-922

Try here: https://studio.ensembleui.com/app/yeveBgUUti3DDVe8X3DE/screen/z9VD6YRNWqQ55GDL7ekv

sagardspeed2:
## Describe your changes
fix paste data on mask pattern update value

Closes #878 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Form/TextInput.tsx
+++ b/packages/runtime/src/widgets/Form/TextInput.tsx
@@ -83,7 +83,7 @@ export const TextInput: React.FC<TextInputProps> = (props) => {
 
   const handleInputPaste = useCallback(
     (e: React.ClipboardEvent) => {
-      const pastedData = e.clipboardData.getData("Text");
+      const pastedData = e.clipboardData.getData("text");
       if (mask) {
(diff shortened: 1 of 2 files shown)


packages/runtime/src/widgets/Form/TextInput.tsx
```diff
@@ -80,6 +81,17 @@ export const TextInput: React.FC<TextInputProps> = (props) => {
     target.value = target.value.replace(/[^0-9.]/g, "");
   }, []);
 
+  const handleInputPaste = useCallback(
+    (e: React.ClipboardEvent) => {
+      const pastedData = e.clipboardData.getData("Text");
```
evshi, line 86: This should be lower case `text`? According to the standard https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-getdata

evshi (approved):

PR #886, opened 2024-10-25
fix: added context in item-template

### LINKED ISSUE #885: No Context in item-template of Collapsible widget
**Describe the bug**
Use `item-template` in Collapsible widget. Example YAML
```yaml
Collapsible:
  item-template:
    data: [1, 2, 3]
    name: p
    template:
      CollapsibleItem:
        key: ${p}
        label: ${'hello' + getDateLabel(p)}
        children:
          Text:
            text: ${'world' + p + ensemble.storage.get("iconName")}
```

**Errors**
image
image
fyi: getDateLabel is available in common.js script in kitchen-sink app

sagardspeed2:
## Describe your changes
Added context in item-template of collapsible widget

### Screenshots [Optional]
After

image

Closes #885 

- [ ] I have added tests


packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -72,6 +85,10 @@ export type CollapsibleProps = {
   HasItemTemplate;
 
 export const Collapsible: React.FC<CollapsibleProps> = (props) => {
+  const screenContext = useScreenContext();
```
evshi, line 88: Do not use screen context directly. Does the deprecated warning not show up for you?

evshi (changes requested):

packages/runtime/src/widgets/Collapsible.tsx
sagardspeed2: it did, but without this, not able to evaluate the value. I take reference from the chart widget for this

evshi: Use the useEvaluate hook for this.

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -106,49 +119,21 @@ export const Collapsible: React.FC<CollapsibleProps> = (props) => {
...
-            // eslint-disable-next-line prefer-named-capture-group
-            .replace(/['"]\$\{([^}]*)\}['"]/g, "$1"), // replace "${...}" or '${...}' with ...
-          {
-            [itemTemplate.name]: get(item, itemTemplate.name) as unknown,
-          },
-        );
-
         return {
           ...item,
-          key: evaluatedConfig.key,
```
evshi, line 127: Where did the key go? This is important for opening programmatically.

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,6 +70,20 @@ export type CollapsibleProps = {
...
 
+const CollapsibleContentRenderer = ({ content }: { content: unknown }) => {
+  const { evaluatedContent } = useEvaluate({ evaluatedContent: content });
+
+  return (
+    <>
+      {isString(content)
+        ? evaluatedContent
+        : EnsembleRuntime.render([
+            unwrapWidget(content as { [key: string]: unknown }),
```
evshi, line 81: This needs to be memoized it can't be called multiple times.

evshi (changes requested):
Please write a test suite to avoid regressions

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +76,76 @@ export type CollapsibleProps = {
...
+  const evaluatedItem = useMemo(() => {
+    const { key, label, children } = evaluated as {
+      [key: string]: Expression<string>;
+    };
+
+    const header = isString(label)
+      ? label
+      : EnsembleRuntime.render([unwrapWidget(label)]);
+
+    const content = isString(children)
```
evshi, line 100: The content can not be a string.

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +76,76 @@ export type CollapsibleProps = {
...
+      ? children
+      : EnsembleRuntime.render([unwrapWidget(children)]);
+
+    return {
+      key,
+      label: header as Expression<string>,
+      children: content as Expression<string>,
```
evshi, line 107: These should be ReactNodes.

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +76,76 @@ export type CollapsibleProps = {
 } & EnsembleWidgetProps &
   HasItemTemplate;
 
+const Panel = ({
```
evshi, line 79: Rename this.

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +76,76 @@ export type CollapsibleProps = {
...
+  scope,
+  setTemplateItems,
+  template,
+}: {
+  scope: CustomScope;
+  template: CollapsibleItem;
+  setTemplateItems: React.Dispatch<React.SetStateAction<CollapsibleItem[]>>;
+}): null => {
+  const evaluated = useEvaluate({ ...template }, { context: scope });
+  const itemAddedRef = useRef(0);
```
evshi, line 89: Why do you need this?

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +76,76 @@ export type CollapsibleProps = {
...
+
+  const evaluatedItem = useMemo(() => {
+    const { key, label, children } = evaluated as {
+      [key: string]: Expression<string>;
```
evshi, line 93: This cast shouldn't be needed. The object type is inferred based on what you pass in.

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +76,76 @@ export type CollapsibleProps = {
...
+      label: header as Expression<string>,
+      children: content as Expression<string>,
+    };
+  }, [evaluated]) as CollapsibleItem;
+
+  useEffect(() => {
+    itemAddedRef.current += 1;
+    if (itemAddedRef.current === 2) {
+      setTemplateItems((prev) => [...prev, { ...evaluatedItem }]);
+    }
```
evshi, line 115: If `itemAddedRef` is being used to track render count, do not do that. Update this hook to check if the item has been previously added based on the key.

apps/kitchen-sink/src/ensemble/screens/home.yaml
```diff
@@ -370,6 +371,19 @@ View:
                     executeCode: |
                       console.log('>>>', {tooltipState})
 
+        - Collapsible:
+            value: [2]
+            item-template:
+              data: [1, 2, 3]
+              name: p
+              template:
+                CollapsibleItem:
```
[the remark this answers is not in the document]
sagardspeed2: We are supporting Collapsible and Accordion as Collapsible widget in react to give backward compatibility for atlas

packages/runtime/src/widgets/Collapsible.tsx
sagardspeed2: what if the user passed the string as children of collapsible item ?
For ex.

```yaml
Collapsible:
  value: [2]
  item-template:
    data: [1, 2, 3]
    name: p
    template:
      CollapsibleItem:
        key: ${p}
        label: ${'hello' + getDateLabel(p)}
        children: hiii there ${p}
```

packages/runtime/src/widgets/Collapsible.tsx
sagardspeed2: I got it, but what is this header and content are string only, the example I give above ?

packages/runtime/src/widgets/Collapsible.tsx
sagardspeed2: 👍

packages/runtime/src/widgets/Collapsible.tsx
evshi: Then that is invalid and an exception will be thrown.

apps/kitchen-sink/src/ensemble/screens/home.yaml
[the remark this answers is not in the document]
evshi: Yes but there is no `CollapsibleItem` in flutter.

packages/runtime/src/widgets/Collapsible.tsx
evshi: Strings are also React nodes.

apps/kitchen-sink/src/ensemble/screens/home.yaml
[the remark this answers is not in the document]
sagardspeed2: Yes I know and also we are not using anywhere, I got this example reference from Anser, we are only using properties of `template` from itemTemplate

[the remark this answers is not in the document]
evshi: Remove this then and put the key back into itemTemplate?

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +75,63 @@ export type CollapsibleProps = {
...
 
+// Custom hook to evaluate an array of expressions using useEvaluate
+const useEvaluateExpressions = (
+  template: CollapsibleItem,
+  namedData: object[],
+): CollapsibleTemplate[] => {
+  const [evaluatedItems, setEvaluatedItems] = useState<CollapsibleTemplate[]>(
+    [],
+  );
+  const [currentIndex, setCurrentIndex] = useState(-1);
```
evshi, line 86: Do not do this! Why did you no longer make this a component? This is going to be `n` renders of the entire Collapsible component instead of individual smaller renders.

Change this back to a component. Add a check in the parent Collapsible component so that it doesn't render anything until all the keys have been evaluated.

evshi (changes requested):

apps/kitchen-sink/src/ensemble/screens/home.yaml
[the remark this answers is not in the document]
evshi: Will continue to support this since this is being used in the atlas app.

cc: @sharjeelyunus @anserwaseem

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +75,79 @@ export type CollapsibleProps = {
...
+    template,
+    currentIndex,
+    setEvaluated,
+  }: {
+    context: object;
+    currentIndex: number;
+    template: CollapsibleItem;
+    setEvaluated: React.Dispatch<React.SetStateAction<CollapsibleTemplate[]>>;
+  }): null => {
+    const evaluatedData = useEvaluate({ ...template }, { context });
```
evshi, line 90: This should evaluate the key, otherwise you will create additional bindings for label and children attributes you don't need.

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +75,79 @@ export type CollapsibleProps = {
...
+      const content = EnsembleRuntime.render([
+        unwrapWidget(children as string),
+      ]);
+
+      setEvaluated((prev) => {
+        const prevEvaluated = [...prev];
+        prevEvaluated[currentIndex] = { key, label: header, children: content };
+        return prevEvaluated;
+      });
+    }, [currentIndex, evaluatedData, setEvaluated]);
```
evshi, line 108: This should only check for the key.

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +75,79 @@ export type CollapsibleProps = {
...
+      const header = isString(label)
+        ? label
+        : EnsembleRuntime.render([unwrapWidget(label)]);
+
+      const content = EnsembleRuntime.render([
+        unwrapWidget(children as string),
+      ]);
+
+      setEvaluated((prev) => {
+        const prevEvaluated = [...prev];
```
evshi, line 104: Why duplicate the array?

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +75,79 @@ export type CollapsibleProps = {
...
+
+    useEffect(() => {
+      const { key, label, children } = evaluatedData;
+
+      const header = isString(label)
+        ? label
+        : EnsembleRuntime.render([unwrapWidget(label)]);
+
+      const content = EnsembleRuntime.render([
+        unwrapWidget(children as string),
```
evshi, line 100: The children and label are the same across all the items right? This can be unwrapped once and passed in.

packages/runtime/src/widgets/Collapsible.tsx
sagardspeed2: we need a shallow copy because React uses reference equality to detect changes. When you modify the array directly without creating a new reference, React assumes nothing has changed.

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +76,86 @@ export type CollapsibleProps = {
...
+        };
+        return [...prev];
+      });
+    }, [evaluatedData]);
+
+    return null;
+  },
+);
+
+EvaluateItem.displayName = "EvaluateItem";
```
evshi, line 114: Rename this to be `CollapsibleItem`.

packages/runtime/src/widgets/Collapsible.tsx
```diff
@@ -71,9 +76,86 @@ export type CollapsibleProps = {
...
+
+const withValueTemplate = <P extends object>(
```
evshi, line 116: There is no need for this to be a HOC since the logic is not shared anywhere. This is just extra syntax for no value

evshi (approved):

PR #887, opened 2024-10-29
improve performance of navigateUrl action

anserwaseem:

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/hooks/useCommandCallback.ts
+++ b/packages/framework/src/hooks/useCommandCallback.ts
@@ -46,7 +46,7 @@ export const useCommandCallback = <
   R = unknown,
 >(
   command: (evalContext: EnsembleContext, ...args: T) => R,
-  apis: { navigate: NavigateFunction; location: EnsembleLocation },
+  apis: { navigate: NavigateFunction; location?: EnsembleLocation },
   dependencies: unknown[] = [],
   callbackContext?: CallbackContext,
 ): ReturnType<typeof useAtomCallback<R, T>> => {
@@ -163,6 +163,7 @@ export const useCommandCallback = <
             },
             styles: theme.Styles,
             ...mapKeys(theme.Tokens ?? {}, (_, key) => key.toLowerCase()),
+            screenContextModel: screenContext.model,
           },
         });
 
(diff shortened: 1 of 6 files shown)


packages/runtime/src/runtime/hooks/useNavigateUrl.ts
```diff
@@ -1,45 +1,77 @@
...
+          action.inputs,
+          defaultScreenModel,
+          context,
+        );
+      }
 
-    setUrlNavigated(true);
-    return navigate(evaluatedInputs.url, { state: evaluatedInputs.inputs });
-  }, [urlNavigated, evaluatedInputs]);
+      navigate(evaluatedUrl, navigationOptions);
```
evshi, line 70: Let's reuse `navigateUrl` here

packages/framework/src/hooks/useCommandCallback.ts
```diff
@@ -163,6 +163,7 @@ export const useCommandCallback = <
             },
             styles: theme.Styles,
             ...mapKeys(theme.Tokens ?? {}, (_, key) => key.toLowerCase()),
+            screenContextModel: screenContext.model,
```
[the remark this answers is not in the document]
anserwaseem: `defaultScreenContext` has the model as undefined, we need `global` and `importedScripts` block from the model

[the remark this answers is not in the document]
evshi: Ah then just use `useScreenModel` where it's needed then. It shouldn't cause any re-renders.

packages/framework/src/shared/common.ts
```diff
@@ -159,3 +159,40 @@ export const error = (value: unknown): void => {
   // eslint-disable-next-line no-console
   console.error(value);
 };
+
+export const expressionReplacer =
```
evshi, line 163: nit: rename to `replace`

packages/framework/src/shared/common.ts
```diff
@@ -159,3 +159,40 @@ export const error = (value: unknown): void => {
...
+  (evaluate: (expr: string) => string) =>
```
evshi, line 164: this can be named to `replacer`

PR #898, opened 2024-11-07
expose selectedLabel prop in Search component

anserwaseem:
## Describe your changes
UseCase:
allow user to pass a custom widget for selected option

### Screenshots [Optional]

https://github.com/user-attachments/assets/9affdb46-91fc-4b85-8ff5-dad2b9cf24e8

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Search.tsx
+++ b/packages/runtime/src/widgets/Search.tsx
@@ -28,7 +28,7 @@ const widgetName = "Search";
 export type SearchProps = {
   placeholder?: string;
   searchKey?: string;
-  searchWidget?: { [key: string]: unknown };
+  selectedLabel?: { [key: string]: unknown };
   onSearch?: {
     debounceMs: number;
   } & EnsembleAction;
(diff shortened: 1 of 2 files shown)


packages/runtime/src/widgets/Search.tsx
```diff
@@ -28,6 +28,7 @@ const widgetName = "Search";
 export type SearchProps = {
   placeholder?: string;
   searchKey?: string;
+  searchWidget?: { [key: string]: unknown };
```
evshi, line 31: I'm not sure if `searchWidget` is the right naming for this property...this should be like resultWidget maybe?

evshi (approved):
In retrospect, I'm not sure why the widget being used in your video is not something like Select/Dropdown with search? This pattern seems to be blend the two use cases together.

anserwaseem:
> In retrospect, I'm not sure why the widget being used in your video is not something like Select/Dropdown with search? This pattern seems to be blend the two use cases together.

Dropdown widget does have autoComplete prop but lacks some properties like debounceMs, ability to clear the selected value and custom selected widget itself

PR #904, opened 2024-11-11
Saving files locally in both yaml and json format.

TayyabAsghar:

Closes #

- [ ] I have added tests


packages/js-commons/src/constants.ts
evshi, line 1: Move this into local-files.

packages/js-commons/src/dto.ts
```diff
@@ -19,28 +21,19 @@ export interface EnsembleDocument {
   };
 }
 
-export enum EnsembleDocumentType {
```
evshi, line 22: Please keep this here don't move it.

packages/js-commons/src/enums.ts
evshi, line 1: Please keep these inside the dto.ts file. It is better organization to create specific module files than to make generic files like `constants.ts` and `enums.ts` and `utils.ts`. It is difficult to manage when everything is generically labeled. Code that is referenced should live next to each other so it's very clear when something is meant to be accessed or kept inside that module.

packages/js-commons/src/dto.ts
```diff
@@ -53,11 +46,10 @@ export interface ApplicationDTO
...
 
   readonly screens: ScreenDTO[];
-  readonly widgets: WidgetDTO[];
-  readonly scripts: ScriptDTO[];
+  readonly widgets?: WidgetDTO[];
+  readonly scripts?: ScriptDTO[];
   readonly theme?: ThemeDTO;
-  readonly themes?: ThemeDTO[];
-  readonly languages?: LanguageDTO[];
+  readonly translations?: LanguageDTO[];
```
evshi, line 52: Rename the DTO if you are going to rename this attribute.

packages/js-commons/src/firebase.ts
```diff
@@ -208,12 +208,3 @@ const getArtifacts = async (
 const QUERY_FILTERS = {
   notArchived: where("isArchived", "!=", true),
 };
-
-export enum CollectionsName {
```
evshi, line 212: Keep this here.

packages/js-commons/src/local-files.ts
```diff
@@ -0,0 +1,262 @@
...
+  ApplicationMetaDTO,
+  EnsembleDocument,
+} from "./dto";
+import { EnsembleDocumentType } from "./enums";
+
+export const getLocalApplicationTransporter = (
+  hiddenEnsembleFolder: string,
+): LocalApplicationTransporter => ({
+  get: async (appId: string): Promise<ApplicationDTO | null> => {
+    const userAppFolder = join(hiddenEnsembleFolder, appId);
```
evshi, line 32: The user app folder should not be inside the hidden ensemble folder. The hidden ensemble folder keeps track of the app paths on the local FS.

packages/js-commons/src/transporter.ts
```diff
@@ -4,3 +4,17 @@ export interface ApplicationTransporter {
   get: (appId: string) => Promise<ApplicationDTO>;
   put: (app: ApplicationDTO, userId: string) => Promise<ApplicationDTO>;
 }
+
+export interface LocalApplicationTransporter {
```
evshi, line 8: This should extend ApplicationTransporter.

packages/js-commons/src/local-files.ts
```diff
@@ -0,0 +1,262 @@
...
+export const getYamlApplicationTransporter = (
+  yamlFolder: string,
+  globalFolder: string,
+): YamlApplicationTransporter => ({
+  get: async (
+    appId: string,
+    documentId: string,
+    documentType: string,
+  ): Promise<string> => {
+    const appsMetaData = await getAppMetaData(globalFolder);
```
evshi, line 110: App metadata should live next to YAML files.

evshi (changes requested):
This is not what we discussed. Please make changes such that:

1. The global hidden ensemble folder keeps track of which apps exist on the local file system and where to find them, i.e. the paths
2. The app metadata is stored in the same folder as the YAML sub directories.

Do not flatten the code organization into generic files, e.g. `enums`, `utilities`, `constants`. Keep code organized in vertical modules that correspond to a specific function, e.g. `local` - node, `dto` - models, `firebase` - cloud

packages/js-commons/src/local-files.ts
TayyabAsghar: This is for the Global Apps MetaData as we discussed. A global single file that holds generic metadata of each app and location where the YAML and JSON files of each app is located.

packages/js-commons/src/local-files.ts
TayyabAsghar: The structure is like that :
1- A hidden global folder that contains a file that holds metadata of all apps and yaml folder paths on user local disk.
2- Each app is downloaded in a separate folder with the app name. Each such app folder contains a hidden folder with name `.ensemble` and a yaml folder for each type of artifact `Screens`, `Widgets`,  `Theme`, `Scripts` and `Translations`.
3- Each Yaml folder contains yaml files with the name of artifact and a hidden `.manifest.json` file to keep track of names of files with the artifact id.

This is what we discussed and this is what we agreed on. Let me know if I missed anything.

packages/js-commons/src/local-files.ts
```diff
@@ -0,0 +1,380 @@
...
+      assets,
+      widgets,
+      scripts,
+      name: "",
+      id: appId,
+      translations,
+      screens: screens ?? [],
+      groupLabels: new Map(Object.entries(groupLabels ?? {})),
+    };
+    return applicationData;
```
evshi, line 93: This doesn't contain the YAML content like we discussed?

packages/js-commons/src/local-files.ts
```diff
@@ -0,0 +1,380 @@
...
+      writeJsonData(jsonFilesPath.env, appData.env ?? {}),
+      writeJsonData(jsonFilesPath.theme, appData.theme ?? {}),
+      writeJsonData(jsonFilesPath.assets, appData.assets ?? []),
+      writeJsonData(jsonFilesPath.screens, appData.screens ?? []),
+      writeJsonData(jsonFilesPath.scripts, appData.scripts ?? []),
+      writeJsonData(jsonFilesPath.widgets, appData.widgets ?? []),
+      writeJsonData(jsonFilesPath.groupLabels, appData.groupLabels ?? []),
+      writeJsonData(jsonFilesPath.translations, appData.translations ?? []),
+    ]);
+    return appData;
```
evshi, line 129: This doesn't write the YAML files like we discussed...

packages/js-commons/src/local-files.ts
```diff
@@ -0,0 +1,380 @@
...
+        join(folderPath, HiddenManifestFile),
+        artifactMetaData,
+        true,
+      ),
+    ]);
+};
+
+const getYamlFolderPath = async (
+  appId: string,
+  yamlFolder: string,
```
evshi, line 317: What is this for? Since the app can be anywhere on the local FS I don't see the point of adding another path here?

packages/js-commons/src/local-files.ts
```diff
@@ -0,0 +1,380 @@
...
+      readJsonFile<AssetDTO[]>(join(userAppFolder, ArtifactFilesName.assets)),
+      readJsonFile<ScreenDTO[]>(join(userAppFolder, ArtifactFilesName.screens)),
+      readJsonFile<WidgetDTO[]>(join(userAppFolder, ArtifactFilesName.widgets)),
+      readJsonFile<ScriptDTO[]>(join(userAppFolder, ArtifactFilesName.scripts)),
+      readJsonFile<Record<string, string>>(
+        join(userAppFolder, ArtifactFilesName.groupLabels),
+      ),
+      readJsonFile<TranslationDTO[]>(
+        join(userAppFolder, ArtifactFilesName.translations),
+      ),
```
evshi, line 78: Why did you decide to break this metadata up into individual subdirectories? This can all be stored in one manifest file at the hidden app folder level. Opening up multiple file handles consumes more memory than necessary.

evshi (changes requested):
Still needs quite a bit of work.

packages/js-commons/src/local-files.ts
TayyabAsghar: Since app can be anywhere on the FS this is the global metadata file to keep track of yaml folder on the fs.

evshi: Then what is the difference between YAMLFolder and GlobalFolder?

TayyabAsghar: The idea was in the future users could update the path of App Yaml files from the web studio if they had moved the app from anywhere. This Yaml Folder path will act as a fallback path if they accidentally remove the Yaml Filepath to save the app data again.

Issue #906, opened 2024-11-14
Clear the Search widget results while clearing its input 


justEhmadSaeed:
**Describe the bug**
when a user enters something in the input field of the Search widget and takes the focus away, it clears the input, but the search results don't get reset. The expected behaviour it to clear both the input text and the search results

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #909, opened 2024-11-18
[906] chore: clear search results when focused outside the Search widget

justEhmadSaeed:
## Describe your changes
#906 
### Screenshots [Optional]

https://github.com/user-attachments/assets/e8ae35d2-2de8-4106-8349-c61316413a58

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Search.tsx
+++ b/packages/runtime/src/widgets/Search.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState, useMemo } from "react";
+import React, { useCallback, useState, useMemo, useEffect } from "react";
 import { useDebounce } from "react-use";
 import {
   useTemplateData,
@@ -74,6 +74,12 @@ export const Search: React.FC<SearchProps> = ({
   const onSelectAction = useEnsembleAction(onSelect);
   const onClearAction = useEnsembleAction(onClear);
 
+  useEffect(() => {
+    if (!searchValue) {
+      setHasCleared(true);
+    }
+  }, [searchValue]);
+
   const extractValue = useCallback(
     (option: unknown): string | number => {
       return get(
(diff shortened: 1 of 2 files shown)


packages/runtime/src/widgets/Search.tsx
```diff
@@ -227,6 +227,7 @@ export const Search: React.FC<SearchProps> = ({
           renderLabel(label, labelValue)
         }
         notFoundContent={notFoundContentRenderer}
+        onBlur={handleClear}
```
sharjeelyunus, line 230: shouldn't we expose the onBlur property with EnsembleAction instead of clearing the results.

Clearing means user will lost the previously searched results, this might solve the current problem, but not a general use case

CC: @evshi

evshi: The issue we are trying to solve here is if the value is emptied, then the search results should also be emptied. It doesn't necessarily have anything to do with `blur`. Ticket for context https://atlashealth.atlassian.net/browse/TRINITY-1461

@justEhmadSaeed what is clearing the search input value?

evshi: In that case, we should add an effect/fix the effect that syncs the search results with the value. If the value was set to be null, the search results should get cleared @justEhmadSaeed

justEhmadSaeed: @evshi updated the PR

packages/runtime/src/widgets/Search.tsx
```diff
@@ -74,6 +74,12 @@ export const Search: React.FC<SearchProps> = ({
   const onSelectAction = useEnsembleAction(onSelect);
   const onClearAction = useEnsembleAction(onClear);
 
+  useEffect(() => {
+    if (!searchValue) {
+      setHasCleared(true);
+    }
+  }, [searchValue]);
```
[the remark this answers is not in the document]
justEhmadSaeed: @evshi yeah if we modify the condition to return an empty array if either searchValue is empty or hasCleared is true

evshi (approved):

PR #911, opened 2024-11-20
missing theme context in useCommandCallback

sagardspeed2:
## Describe your changes
Added missing themeContext in useCommandCallback

Closes #

- [ ] I have added tests

### DIFF
--- a/.changeset/bright-walls-reply.md
+++ b/.changeset/bright-walls-reply.md
@@ -1,0 +1,6 @@
+---
+"@ensembleui/react-framework": patch
+"@ensembleui/react-runtime": patch
+---
+
+fix: refactor missing themeContext in useCommandCallback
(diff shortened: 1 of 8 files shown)


sagardspeed2:
> @sagardspeed2 what is this for? The theme is injected as an atom here https://github.com/EnsembleUI/ensemble-react/blob/main/packages/framework/src/hooks/useCommandCallback.ts#L63

@evshi I know that theme is injected as atom, but it missing the method like `setTheme`, and due to that is breaks the functionality related change the theme. so I added themeContext which includes that method

evshi:
@sagardspeed2 add a new atom with the methods and use that instead. Prefer using atoms instead of context hooks.

evshi (approved):

PR #913, opened 2024-11-21
fix: datagrid render issue with namedData

sagardspeed2:
## Describe your changes
fix datagrid infinite render issue with namedData

Closes #

- [ ] I have added tests


packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -289,6 +289,7 @@ export const DataGrid: React.FC<GridProps> = (props) => {
     [curPage, onPageChangeActionCallback],
   );
 
+  const namedDataString = JSON.stringify(namedData);
```
sagardspeed2, line 292: Here `namedData` is a array we are getting from `useTemplateData` and we are passing this in the callback dependancies, and this array is generating new reference address on each render, so it cause the infinite render, because we are passing this method in `useRegisterBindings`.

so I converted this into JSON.stringify, the same we handle for `allowedExtension` in `usePickFiles`

evshi: `useTemplateData` memoizes the array, it shouldn't return a new reference on each render.

evshi: @sagardspeed2 please fix this in the `useTemplateData` hook like we discussed.

packages/framework/src/hooks/useTemplateData.ts
```diff
@@ -21,6 +21,15 @@ export interface TemplateDataProps {
  * @param name - the name each instance of data should be keyed by
  * @returns
  */
+
+const useDeepCompareMemoize = <T>(value: T) => {
+  const ref = useRef<T>();
+  if (!isEqual(value, ref.current)) {
```
evshi, line 27: Use react-fast-compare

evshi (approved):

PR #918, opened 2024-11-22
fix: added cache in API call

### LINKED ISSUE #350: Optimize graphql query state
GraphQL protocol cannot use http cache headers because everything hits the same resource/endpoint. Instead graphQL has its own caching directives, and client can make further caching optimizations.

Use apollo client caching or react-query caching to avoid refetching expensive queries

https://www.apollographql.com/docs/react/caching/overview
https://tanstack.com/query/v4/docs/framework/react/guides/caching

sagardspeed2:
## Describe your changes
fix added cache support in API trigger

Closes #350 

- [ ] I have added tests

### DIFF
--- a/apps/kitchen-sink/src/ensemble/screens/home.yaml
+++ b/apps/kitchen-sink/src/ensemble/screens/home.yaml
@@ -6,6 +6,8 @@ View:
       ensemble.storage.set('zzz' , 'button')
       ensemble.storage.set('aaa' , '1')
       ensemble.storage.set('bbb' , "0xffb91c1c")
+      ensemble.storage.set('email', 'yash@yopmail.com');
+      ensemble.storage.set('emails', ['yash@yopmail.com']);
       console.log('>>> secret variable >>>', ensemble.secrets.dummyOauthSecret)
 
   header:
@@ -370,6 +372,28 @@ View:
                     executeCode: |
                       console.log('>>>', {tooltipState})
 
+        - Text:
+            text: ${ensemble.storage.get('email')}
+        - Button:
+            label: Call Action
+            onTap:
+              invokeAPI:
+                name: getDummyProducts
+                onResponse:
+                  executeCode: |
+                    ensemble.storage.set('email', response.body.results[0].email)
+                    ensemble.storage.set('emails', [...ensemble.storage.get("emails"),response.body.results[0].email])
+                    console.log('getData', response.body.results[0].email, ensemble.storage.get('emails'));
+        - Column:
+            item-template:
+              data: '${ensemble.storage.get("emails")}'
+              name: email
+              template:
+                Column:
+                  children:
+                    - Text:
+                        text: ${email}
+
 Global:
   scriptName: test.js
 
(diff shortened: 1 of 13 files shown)


packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -208,32 +210,37 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
...
-              env: appContext?.env,
-              secrets: appContext?.secrets,
-            },
-          },
-          {
-            mockResponse: mockResponse(
-              mockResponses[api.name],
-              useMockResponse,
+        const response = await queryClient.fetchQuery({
+          queryKey: [api.name],
```
evshi, line 214: This needs to be a hash of screen/name/inputs. Otherwise the invoking the same API again on a different screen or different inputs will return the wrong result.

Add a test for this.

apps/kitchen-sink/src/ensemble/screens/home.yaml
```diff
@@ -396,3 +420,9 @@ API:
   ERROR500:
     method: GET
     uri: https://httpstat.us/500
+
+  getDummyProducts:
+    method: GET
+    cache: true
+    cacheTime: 10
```
evshi, line 427: Consolidate to one property called `cacheExpiry`

evshi (changes requested):
You also need to add caching to the other invocation for executeCode.

packages/framework/src/api/data.ts
```diff
@@ -11,6 +12,8 @@ import { screenDataAtom, type ScreenContextDefinition } from "../state";
 import { isUsingMockResponse } from "../appConfig";
 import { mockResponse } from "../evaluate/mock";
 
+const queryClient = new QueryClient();
```
evshi, line 15: You need to use the same instance of the queryClient otherwise it will not be the same cache.

packages/framework/src/api/data.ts
```diff
@@ -23,6 +26,12 @@ export const invokeAPI = async (
     (model) => model.name === apiName,
   );
 
+  const hash = JSON.stringify({
```
evshi, line 29: Make a re-usable function for this that can be shared by framework and runtime.

packages/runtime/src/runtime/hooks/__tests__/useActionGroup.test.tsx
```diff
@@ -164,6 +190,120 @@ test("fetch multiple APIs", async () => {
   });
 });
 
+test("fetch API cache response for cache expiry", async () => {
```
evshi, line 193: Move these tests to the useInvokeApi test suite. Also add a test that asserts if the cacheExpiry is not present, then the response is not cached.

packages/framework/src/shared/models.ts
```diff
@@ -70,6 +70,7 @@ export interface EnsembleAPIModel {
   url?: string;
   method: "GET" | "POST" | "PUT" | "PATCH";
   headers?: { [key: string]: string | number | boolean };
+  cacheExpiry?: number;
```
evshi, line 73: Let's rename this to `cacheExpirySeconds` to make it clear what the unit of measurement is.

evshi (changes requested):

evshi (approved):

Issue #921, opened 2024-11-25
Allow default value for the Search widget


justEhmadSaeed:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #922, opened 2024-11-25
921 allow default value for search widget

justEhmadSaeed:
## Describe your changes
#921 

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Search.tsx
+++ b/packages/runtime/src/widgets/Search.tsx
@@ -26,6 +26,7 @@ import { Icon } from "./Icon";
 const widgetName = "Search";
 
 export type SearchProps = {
+  value?: string;
   placeholder?: string;
   searchKey?: string;
   selectedLabel?: { [key: string]: unknown };
@@ -50,6 +51,7 @@ export const Search: React.FC<SearchProps> = ({
   onChange,
   onSelect,
   onClear,
+  value: initialValue,
   ...rest
 }) => {
   const [searchValue, setSearchValue] = useState<string | null>(null);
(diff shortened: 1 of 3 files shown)


packages/runtime/src/widgets/Search.tsx
```diff
@@ -61,7 +63,7 @@ export const Search: React.FC<SearchProps> = ({
   });
 
   const { id, rootRef, values } = useRegisterBindings(
-    { styles, value, ...rest, widgetName },
+    { styles, value, ...rest, widgetName, initialValue },
```
evshi, line 66: Subtle but important distinction - do you also want the `value` in widget state to be initialized to initialValue? This will update the search results, allow it to be referenced via variable, etc. Basically, is this only a visual indicator?

If so, you will need to add a hook to update the local component `value` state variable.

If not, then you should make this a separate prop called `initialValue` to avoid confusion.

justEhmadSaeed: @evshi i'm intending on using it like the default `value` of the `Dropdown`

evshi: @justEhmadSaeed then you need to add the useEffect hook that syncs the initialValue to the widget state value.

justEhmadSaeed: @evshi updated the PR to sync the widget state value with that of the initialValue

packages/runtime/src/widgets/Search.tsx
```diff
@@ -183,6 +185,10 @@ export const Search: React.FC<SearchProps> = ({
     [extractValue, namedData, rest.selectedLabel],
   );
 
+  useEffect(() => {
+    setValue(values?.initialValue);
+  }, [values?.initialValue]);
```
evshi, line 190: This should only set the value if there is no existing value already.

justEhmadSaeed: @evshi there will be no value on first render every time and due to initialValue being a constant string, it won't run again.
but if i add a check for `value` in it, it will check it every time the `value` gets updated. correct me if im wrong

evshi (approved):

PR #923, opened 2024-11-27
fix: add useCommandCallback in actions

### LINKED ISSUE #846: Refactor imperative actions to use `useCommandCallback`
The old way we implemented imperative actions (useExecuteCode, navigate, etc.) relied on re-renders to get the latest values. `useCommandCallback` defers the evaluation of the atoms to when the command is invoked and doesn't re-render the component on each dependency update.

sagardspeed2:
## Describe your changes
Added useCommandCallback in actions

### Screenshots [Optional]
Before, where test case has been failed
image

After fix in formatJS of evaluate, test cases has been passed
image

Closes #846 

- [ ] I have added tests


packages/runtime/src/runtime/hooks/useNavigateModal.tsx
```diff
@@ -50,40 +65,29 @@ export const useNavigateModalScreen: EnsembleActionHook<
...
+        onDismissCallback,
+      );
+    },
+    { navigate },
+    [action, applicationContext?.application, onDismissCallback, screenModel],
+  );
 
-    return { callback };
-  }, [action]);
+  return useMemo(() => ({ callback: navigateCommand }), [navigateCommand]);
```
evshi, line 92: I thought we agreed we would have the components memoize the callback so this is not needed?

packages/runtime/src/runtime/hooks/useNavigateModal.tsx
```diff
@@ -18,21 +26,28 @@ import {
...
     !isString(action) && action ? action.onModalDismiss : undefined,
   );
-  const [context, setContext] = useState<unknown>();
-  const [isComplete, setIsComplete] = useState<boolean>();
 
-  const evaluatedInputs = useEvaluate(
-    {
-      inputs:
-        !isString(action) && action?.inputs ? cloneDeep(action.inputs) : {},
+  const evaluateInputs = useCallback(
```
evshi, line 38: This doesn't need to be callback this can just be re-usable shared function.

packages/runtime/src/runtime/hooks/useNavigateModal.tsx
```diff
@@ -50,40 +65,29 @@ export const useNavigateModalScreen: EnsembleActionHook<
...
-      setIsComplete(false);
-      setContext(args);
-    };
+      navigateModalScreen(
+        action,
+        modalContext,
+        applicationContext?.application,
+        evaluatedInputs,
+        title,
+        onDismissCallback,
```
evshi, line 85: We can get rid of the extra hook and just pass the ensembleAction callback directly.

packages/runtime/src/runtime/hooks/useNavigateModal.tsx
sagardspeed2: components memoize the callback but we need to return the memoized callback from each action also, that we did in all the actions when we worked on the method stale PR

evshi: `useCommandCalllback` already returns a memoized callback so this is not necessary.

packages/runtime/src/shared/common.ts
```diff
@@ -0,0 +1,18 @@
+import type { EnsembleScreenModel } from "@ensembleui/react-framework";
+import {
+  evaluate,
+  replace,
+  visitExpressions,
+} from "@ensembleui/react-framework";
+
+export const evaluateInputs = (
```
evshi, line 8: Put this in framework. Rename to `evaluateDeep` so it is more general and rename variables to reflect more general use cases.

evshi (changes requested):

packages/runtime/src/runtime/hooks/useNavigateModal.tsx
sagardspeed2: @evshi you are right that `useCommandCalllback` already returns a useCallback, but we are returning new object from the action and to memoize that we need this, if you check all other actions of ensembleActions, we are doing same

` return { callback: navigateCommand };`

please refer useNavigateUrl, useExecuteCode, useCommandCallback, useCloseAllDialogs and so on

packages/runtime/src/shared/common.ts
sagardspeed2: I already tried this by adding `evaluateInputs` into common.ts of framework, but it will create cycling dependancy, and only for that I created new common in runtime.

we are importing evaluate from evaluate.ts, which also import other methods from common.ts, which will create the cycling dependancy

packages/runtime/src/runtime/hooks/useNavigateScreen.ts
```diff
@@ -1,57 +1,67 @@
...
+  const screenModel = useScreenModel();
   const appContext = useApplicationContext();
-  const [context, setContext] = useState<unknown>();
-  const [isComplete, setIsComplete] = useState<boolean>();
 
-  const evaluatedInputs = useEvaluate(
-    {
-      inputs: hasOptions ? cloneDeep(action?.inputs) : null,
-      name: hasOptions ? action?.name : action,
+  const evaluateScreenName = useCallback(
```
evshi, line 24: This hook isn't doing anything.

packages/runtime/src/runtime/hooks/useNavigateModal.tsx
```diff
@@ -18,29 +21,18 @@ import {
...
-  );
-
-  const onDismissCallback = useCallback(() => {
-    if (!ensembleAction) {
-      return;
-    }
-    ensembleAction.callback();
-  }, [ensembleAction]);
+  const onClose = (): void => {
+    ensembleAction?.callback();
```
evshi, line 34: You can pass the callback directly without wrapping it.

packages/runtime/src/runtime/hooks/useNavigateModal.tsx
evshi: @sagardspeed2 that is what my original comment is referring to. We discussed last week that it is not necessary to memoize the return object because the consuming component should only depend on the callback instead of the object.

sagardspeed2:
@evshi check these

Before fix in formatJS, where test cases has been failed
image


After fix in formatJS of evaluate, test cases has been passed
image

evshi (approved):

PR #935, opened 2024-12-03
fix: added support for onResponse and onError in ensemble.invokeAPI

### LINKED ISSUE #860: API onResponse and onError do not execute when invoking API via code
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

sagardspeed2:
added support for onResponse and onError in ensemble.invokeAPI

Closes #860 

- [ ] I have added tests


packages/framework/src/shared/models.ts
```diff
@@ -72,8 +72,8 @@ export interface EnsembleAPIModel {
   headers?: { [key: string]: string | number | boolean };
   cacheExpirySeconds?: number;
   body?: string | object;
-  onResponse?: EnsembleAction;
-  onError?: EnsembleAction;
+  onResponse?: EnsembleAction | EnsembleActionHookResult;
+  onError?: EnsembleAction | EnsembleActionHookResult;
```
evshi, line 76: Store as separate property

packages/runtime/src/runtime/screen.tsx
```diff
@@ -34,6 +35,7 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
   const routeParams = useParams(); // route params
   const params = new URLSearchParams(search); // query params
   const queryParams: { [key: string]: unknown } = Object.fromEntries(params);
+  const [apis, setApis] = useState<EnsembleAPIModel[]>([]);
```
evshi, line 38: Use a map for easy lookup.

packages/runtime/src/runtime/screen.tsx
```diff
@@ -91,34 +93,76 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
...
+  api,
+  currentIndex,
+  setApis,
+}: {
+  api: EnsembleAPIModel;
+  currentIndex: number;
+  setApis: React.Dispatch<React.SetStateAction<EnsembleAPIModel[]>>;
+}): null => {
+  const onResponse = useEnsembleAction(api.onResponse as EnsembleAction);
+  const onError = useEnsembleAction(api.onError as EnsembleAction);
```
evshi, line 154: This needs to happen inside the ScreenProvider/ModalWrapper.

sagardspeed2: I think this is the best place to convert that into callback, because we need to this in runtime only. other thing is to after converting them into callback, we need to store them in screenAtom, thus we can consume that into framework by accessing from screenAtom by screenContext.

packages/runtime/src/runtime/screen.tsx
sagardspeed2: @evshi sorry, but I did not understand your suggestion about use of map

packages/runtime/src/runtime/screen.tsx
evshi: I don't think you are understanding me here. This evaluation needs to happen inside of the screen and modal context otherwise the other actions won't work

packages/runtime/src/runtime/screen.tsx
evshi: Instead of an array of API models, use a map.

packages/runtime/src/runtime/screen/wrapper.tsx
```diff
@@ -0,0 +1,43 @@
...
+  children,
+}) => {
+  const { apis = [], setApi } = useScreenData();
+
+  return (
+    <>
+      {apis.map((api, index) => (
+        <EvaluateApi
+          api={api}
+          currentIndex={index}
```
evshi, line 16: You don't need index. Each API in the screen is uniquely identified by its name.

packages/runtime/src/runtime/screen/index.tsx
```diff
@@ -102,18 +103,20 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
...
+          </OnLoadAction>
+          {screen.menu ? (
+            <EnsembleMenu
+              menu={{ ...screen.menu }}
+              renderOutlet={false}
+              type={screen.menu.type}
+            />
+          ) : null}
+          <EnsembleFooter footer={screen.footer} />
+        </ScreenApiWrapper>
```
evshi, line 119: This doesn't need to wrap the children elements it can be a sibling.

evshi (approved):

PR #937, opened 2024-12-03
add asset storage functionality offline

anserwaseem:

Closes #

- [ ] I have added tests


packages/js-commons/src/local-files.ts
```diff
@@ -68,7 +68,13 @@ export const getLocalApplicationTransporter = (
           } else {
             return;
           }
-          const content = await readFile(filePath, { encoding: "utf-8" });
```
evshi, line 71: `filePath` here refers to the location of the YAML metadata file. The actual asset file location will be either have to be `AssetDTO.publicUrl` or a new attribute that you add.

packages/js-commons/src/local-files.ts
```diff
@@ -68,7 +68,13 @@ export const getLocalApplicationTransporter = (
...
+
+          let content = "";
+          if (document.type === EnsembleDocumentType.Asset) {
+            const binaryContent = await readFile(filePath);
+            content = binaryContent.toString("base64");
```
evshi, line 75: You can't assume it is stored in base64.

packages/js-commons/src/local-files.ts
```diff
@@ -201,6 +207,47 @@ export const saveArtifact = async (
...
+    throw new Error(`App ${appId} not found in local metadata`);
+  }
+
+  const assetDir = join(
+    appMetadata.projectPath,
+    isFont ? EnsembleDocumentType.Font : EnsembleDocumentType.Asset,
+  );
+  ensureDir(assetDir);
+
+  const assetPath = join(assetDir, fileName);
```
evshi, line 230: Again, need to separate the file metadata from the file itself.

packages/js-commons/src/local-files.ts
```diff
@@ -201,6 +207,47 @@ export const saveArtifact = async (
   return { relativePath: pathToWrite };
 };
 
+export const storeAsset = async (
+  appId: string,
+  fileName: string,
+  fileData: string,
```
evshi, line 213: This should be a binary array/buffer

packages/js-commons/src/local-files.ts
```diff
@@ -201,6 +207,47 @@ export const saveArtifact = async (
...
+  }
+
+  const assetDir = join(
+    appMetadata.projectPath,
+    isFont ? EnsembleDocumentType.Font : EnsembleDocumentType.Asset,
+  );
+  ensureDir(assetDir);
+
+  const assetPath = join(assetDir, fileName);
+  const binaryData = Buffer.from(fileData, "base64");
```
evshi, line 231: There's no need to convert to/from base64.

evshi (changes requested):

evshi (commented):
Can you also add:

1. a utility to read the file - this is going to be needed by studio to serve the asset for preview
2. in `put` for both local and firebase apps, handle writing the file pointed to by the metadata document to the file system or firebase, respectively.

packages/js-commons/src/core/local-files.ts
```diff
@@ -167,8 +177,10 @@ export const saveArtifact = async (
   options: {
     relativePath?: string;
     skipMetadata?: boolean;
+    skipContent?: boolean;
   } = {
     skipMetadata: false,
+    skipContent: false,
```
evshi, line 183: Don't add this. This function is only for writing the artifact. Create a separate function for saving the file contents to the system.

evshi (changes requested):
Please separate the actual file from the ensemble artifact that describes it.

packages/js-commons/src/core/local-files.ts
evshi: If the file and the artifact are separate writes then this shouldn't happen. The EnsembleDocument of Asset/Font has no content just a reference to the location of the actual file.

packages/js-commons/src/core/local-files.ts
```diff
@@ -188,15 +286,31 @@ export const saveArtifact = async (
     pathToWrite = `${artifact.name}.yaml`;
   }
 
-  await writeFile(join(artifactSubDir, pathToWrite), artifact.content, "utf-8");
+  if (
+    artifact.type !== EnsembleDocumentType.Asset &&
+    artifact.type !== EnsembleDocumentType.Font
+  ) {
```
evshi, line 292: Don't need this anymore?

packages/js-commons/src/core/transporter.ts
```diff
@@ -3,6 +3,21 @@ import type { ApplicationDTO } from "./dto";
 export interface ApplicationTransporter {
   get: (appId: string) => Promise<ApplicationDTO>;
   put: (app: ApplicationDTO, userId: string) => Promise<ApplicationDTO>;
+  storeAsset: (
```
evshi, line 6: These don't need to part of the interface they can just be exported as utilities.

packages/js-commons/src/core/local-files.ts
```diff
@@ -158,6 +173,89 @@ export const getLocalApplicationTransporter = (
...
+
+    storeAsset: async (
+      appId: string,
+      fileName: string,
+      fileData: string | Buffer,
+      isFont = false,
+      fontFamily?: string,
+      weight?: number,
+      fontStyle?: string,
+      fontType?: string,
```
evshi, line 185: Combine this into an optional object property, e.g. `font?: { family, weight, etc..}`

packages/js-commons/src/core/local-files.ts
anserwaseem: we still do, writing asset & font is dealt in `localStoreAsset` method, `saveArtifact` is called only for non-asset artifacts

packages/js-commons/src/core/local-files.ts
```diff
@@ -97,70 +103,173 @@ export const getLocalApplicationTransporter = (
...
-          existingAppMetadata,
-          {
+        let relativePath = "";
+        let docToWrite = document;
+
+        if (isAssetOrFont(document)) {
+          const asset = document as AssetDTO | FontDTO;
+          if (!asset.publicUrl) return;
+
+          const fileData = await fetchFileData(asset.publicUrl);
```
evshi, line 143: You will need to consider that sometimes the asset needs to be fetched from the local file system, and not the publicUrl.

evshi (approved):

PR #967, opened 2024-12-27
add method for deleting ensemble app locally

anserwaseem:

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/js-commons/src/core/local-files.ts
+++ b/packages/js-commons/src/core/local-files.ts
@@ -1,7 +1,7 @@
 import { join } from "node:path";
 import { exec } from "node:child_process";
 import { homedir } from "node:os";
-import { existsSync, mkdirSync, unlinkSync } from "node:fs";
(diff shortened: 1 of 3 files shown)


packages/js-commons/src/core/local-files.ts
```diff
@@ -184,6 +184,24 @@ export const getLocalApplicationTransporter = (
...
     },
+
+    delete: async (appId: string): Promise<void> => {
+      const appsMetaData = await getGlobalMetadata();
+      const existingAppMetadata = appsMetaData[appId];
+      if (!existingAppMetadata) {
+        throw Error(`App ${appId} not found on local disk`);
+      }
+      // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+      delete appsMetaData[appId];
```
evshi, line 195: This should happen only after the directory is removed. Otherwise if something goes wrong with deleting the directory you will have lost the pointer.

evshi (changes requested):

evshi (approved):

PR #968, opened 2024-12-27
fix: improve regex of replacer in evaluateDeep

sagardspeed2:
## Describe your changes
Improve regex of replacer in evaluateDeep

The current regex is not capable to evaluate `${ensemble.user.userDetails?.clients?.map(({id}) => id) ?? []}`

Closes #

- [x] I have added tests

### DIFF
--- a/packages/framework/src/shared/__tests__/common.test.ts
+++ b/packages/framework/src/shared/__tests__/common.test.ts
@@ -1,3 +1,4 @@
+import { evaluateDeep } from "../../evaluate";
 import { findExpressions, isCompoundExpression } from "../common";
 
 test("find deeply nested expressions", () => {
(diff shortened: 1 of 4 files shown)


packages/framework/src/shared/common.ts
```diff
@@ -163,11 +163,12 @@ export const error = (value: unknown): void => {
 export const replace =
   (replacer: (expr: string) => string) =>
   (val: string): unknown => {
-    const matches = val.match(/\$\{+[^}]+\}+/g);
+    const replaceRegex = /\$\{(?:[^}{]+|\{(?:[^}{]+|\{[^}{]*\})*\})*\}/g;
```
evshi, line 166: How did you come up with this regex? This is going to take anyone a long time to understand. Please leave a comment and tests to describe what this does.

Consider replacing this regex with a stack implementation instead since regex is infamously not meant for capturing things like matching parens/braces.

packages/framework/src/shared/common.ts
```diff
@@ -160,16 +160,55 @@ export const error = (value: unknown): void => {
   console.error(value);
 };
 
+export const validateExpressions = (
+  value: string,
+): { isValid: boolean; expressions: string[] } => {
+  let expression = "";
+  let stake = 0;
```
evshi, line 167: Is this supposed to be `stack`?

evshi (changes requested):
Need to add tests

packages/framework/src/shared/common.ts
```diff
@@ -160,16 +160,61 @@ export const error = (value: unknown): void => {
...
 
+export const validateExpressions = (
+  value: string,
+): { isValid: boolean; expressions: string[] } => {
+  const expressions: string[] = [];
+  let currentExpr = "";
+  let depth = 0;
+  let i = 0;
+
+  while (i < value.length) {
```
evshi, line 171: Why go from for loop to while?

packages/framework/src/shared/common.ts
```diff
@@ -160,16 +160,61 @@ export const error = (value: unknown): void => {
   console.error(value);
 };
 
+export const validateExpressions = (
+  value: string,
+): { isValid: boolean; expressions: string[] } => {
+  const expressions: string[] = [];
+  let currentExpr = "";
+  let depth = 0;
```
evshi, line 168: Rename this.

packages/framework/src/shared/__tests__/common.test.ts
```diff
@@ -82,4 +83,115 @@ describe("isCompoundExpression", () => {
...
+        {
+          name: "${['X', 'Y'].map((c) => { return c.toLowerCase() })}",
+        },
+        undefined,
+        {
+          name: "Ensemble",
+          platform: "Web",
+        },
+      ),
+    ).toMatchObject({ name: ["x", "y"] });
```
evshi, line 194: Add a case for the type of expression that created the issue. As well as tests where the evaluation should fail.

packages/framework/src/data/fetcher.ts
```diff
@@ -35,9 +35,25 @@ class EnsembleResponse implements Response {
...
+    return response;
+  }
+
+  static failApi(): EnsembleResponse {
+    const response = new EnsembleResponse();
+    response.statusCode = undefined;
+    response.isLoading = false;
+    return response;
+  }
 }
```
evshi, line 52: Please get rid of these. You can update the constructor to initialize the attributes for new responses. But for failed states, mutate the existing object.

evshi (changes requested):

evshi (approved):

PR #969, opened 2024-12-27
fix/ensemble-response-mapping

sagardspeed2:
## Describe your changes
fix/ensemble-response-mapping

basically assign was overwriting the Response isSuccess getter method and thus it was giving type error

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/data/index.ts
+++ b/packages/framework/src/data/index.ts
@@ -8,8 +8,8 @@ export interface Response {
   headers?: fetcher.Headers;
   reasonPhrase?: string;
   isLoading: boolean;
-  isSuccess: boolean;
-  isError: boolean;
+  isSuccess?: boolean;
+  isError?: boolean;
 }
 
 export interface WebSocketConnection {
--- a/packages/framework/src/data/index.ts
+++ b/packages/framework/src/data/index.ts
@@ -8,8 +8,8 @@ export interface Response {
   headers?: fetcher.Headers;
   reasonPhrase?: string;
   isLoading: boolean;
-  isSuccess?: boolean;
-  isError?: boolean;
+  isSuccess: boolean;
+  isError: boolean;
 }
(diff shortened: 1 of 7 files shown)


evshi (commented):
If a request is in flight then `isSuccess` should be false/undefined

packages/framework/src/data/index.ts
```diff
@@ -8,8 +8,8 @@ export interface Response {
   headers?: fetcher.Headers;
   reasonPhrase?: string;
   isLoading: boolean;
-  isSuccess: boolean;
-  isError: boolean;
+  isSuccess?: boolean;
+  isError?: boolean;
```
evshi, line 12: Why did this need to be changed?

packages/framework/src/state/screen.ts
```diff
@@ -40,7 +40,7 @@ export const screenDataAtom = atom(
   (get) => get(screenDataFocusAtom),
   (get, set, update) => {
     const currentData = get(screenDataFocusAtom);
-    const nextData = assign({}, currentData, update);
+    const nextData = merge({}, currentData, update);
```
evshi, line 43: This should still be assign?

sagardspeed2: Oops, let me just update that

packages/framework/src/data/index.ts
sagardspeed2: This both needs to be optional as we are now just updating the statusCode on API trigger, and isSuccess and isError will update through statusCode by getters

evshi: You should update the function type to take Partial instead of updating the root interface.

sagardspeed2: @evshi done

evshi (approved):

PR #972, opened 2024-12-31
fix: prevent modalContext override in API response handler

sagardspeed2:
## Describe your changes
prevent modalContext override in API response handler

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
+++ b/packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
@@ -134,7 +134,7 @@ export const useExecuteCode: EnsembleActionHook<
       if (!js) {
         return;
       }
-      const context = merge({}, ...args, options?.context, evalContext) as {
+      const context = merge({}, evalContext, ...args, options?.context) as {
         [key: string]: unknown;
       };
       const retVal = evaluate({ model: screenModel }, js, context);
@@ -175,10 +175,12 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
   const onAPIErrorAction = useEnsembleAction(currentApi?.onError);
 
   const invokeCommand = useCommandCallback(
-    async (evalContext, ...args) => {
+    async (evalContext, ...args: unknown[]) => {
       if (!action?.name || !currentApi) return;
 
-      const context = merge({}, evalContext, args[0]);
+      const context = merge({}, evalContext, ...args) as {
+        [key: string]: unknown;
+      };
 
       if (action.name !== currentApi.name) return;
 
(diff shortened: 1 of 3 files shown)


packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -175,10 +175,12 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
   const onAPIErrorAction = useEnsembleAction(currentApi?.onError);
 
   const invokeCommand = useCommandCallback(
-    async (evalContext, ...args) => {
+    async (evalContext, ...args: unknown[]) => {
       if (!action?.name || !currentApi) return;
 
-      const context = merge({}, evalContext, args[0]);
+      const context = merge({}, evalContext, ...args) as {
```
evshi, line 181: This needs a check to see if it is an object or array. Or change how this is being passed in by how other hooks are calling it.

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -244,8 +246,14 @@ export const useInvokeAPI: EnsembleActionHook<InvokeAPIAction> = (action) => {
           setData(action.id, response);
         }
 
-        onAPIResponseAction?.callback({ ...context, response });
-        onInvokeAPIResponseAction?.callback({ ...context, response });
+        onAPIResponseAction?.callback({
+          ...(args[0] as { [key: string]: unknown }),
```
evshi, line 250: Why would we not continue to pass context here?

sagardspeed2: Here context is the object of all the ensemble methods that we are generating from useCommandCallback for evaluate the value, in the response action callback, we no need that because in that callback, we will again get from the useCammandCallback, we just need to pass the args that we gets as arguments.

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
sagardspeed2: I only did it, I check the previous implementation of invokeAPI through callback and useEffect and see the other hooks like executeCodeAction, and implemented same here

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
evshi: But there may be evaluated context items that need to be passed down?

sagardspeed2: No, because in subsequent callbacks like useExecuteCode, where we evaluate the context using useCommandCallback, passing the context from here would overwrite the evaluated context in those subsequent callbacks, leading to incorrect or unexpected results.

I have already written a test case for the scenario `after receiving the API response, the modal should close`. This test case verifies that, with the new implementation, we can now successfully close the modal using `ensemble.closeAllDialogs()` in response to the API call. Prior to this fix, the modal was not closing as expected in this scenario.

evshi: This change doesn't make sense with the one above where you are taking the args as an array because this is sending the new context as an object...

evshi (approved):

PR #976, opened 2025-01-06
fix: added validation for check unique API name across screen and wid…

sagardspeed2:
## Describe your changes
when screen and widget has same API name, then screen API is overriding widget API, to fix this issue, added validation in parser that check unique API name across screen and widget

### Screenshots [Optional]
image

Closes #

- [x] I have added tests


packages/framework/src/__tests__/parser.test.ts
```diff
@@ -149,3 +149,34 @@ test("parse multiple theme in a single file", () => {
     },
   });
 });
+
+test("parses application with no custom widgets", () => {
```
evshi, line 153: this test name needs to be updated.

evshi (approved):

packages/framework/src/parser.ts
```diff
@@ -101,6 +101,20 @@ export const EnsembleParser = {
...
     }
 
+    screens.forEach((screen) => {
+      if (!("apis" in screen)) return;
+
+      const apiNames = new Set<string>();
+      screen.apis.forEach((api) => {
+        if (apiNames.has(api.name)) {
+          throw new Error(
+            `Application has multiple apis with the same name (${api.name}) on this screen.`,
```
anserwaseem, line 111: instead of "this" screen, can we show screen name in the error as well?

sagardspeed2: Okay, will do that

PR #991, opened 2025-01-16
fix: allow reset options on clear search

sagardspeed2:
## Describe your changes
Allow reset options on clear search in multiselect and search

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Search.tsx
+++ b/packages/runtime/src/widgets/Search.tsx
@@ -119,7 +119,7 @@ export const Search: React.FC<SearchProps> = ({
 
   useDebounce(
     () => {
-      if (onSearchAction?.callback && !isEmpty(searchValue)) {
+      if (onSearchAction?.callback) {
         onSearchAction.callback({ search: searchValue });
       }
     },
--- a/.changeset/four-apples-beg.md
+++ b/.changeset/four-apples-beg.md
@@ -1,0 +1,5 @@
+---
+"@ensembleui/react-runtime": patch
+---
+
+Allow reset options on clear search in multiselect and search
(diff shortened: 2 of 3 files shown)


packages/runtime/src/widgets/Search.tsx
```diff
@@ -119,7 +119,7 @@ export const Search: React.FC<SearchProps> = ({
 
   useDebounce(
     () => {
-      if (onSearchAction?.callback && !isEmpty(searchValue)) {
```
sharjeelyunus, line 122: @sagardspeed2, didn't we add this for a reason previously? here: https://github.com/EnsembleUI/ensemble-react/pull/909/files#diff-c40bba994aece7636430625fad970ec08c0334f7edc12464e796ea3375086432R118

Could you please confirm that removing this won't cause any regressions?

evshi (commented):
Needs tests to avoid regressions like @sharjeelyunus is mentioning

packages/runtime/src/widgets/Form/__tests__/MultiSelect.test.tsx
```diff
@@ -537,6 +537,18 @@ describe("MultiSelect Widget", () => {
...
+
+    // Wait for the combobox to reflect the selected values
+    await waitFor(() => {
+      expect(screen.getByText("Option 4", { selector })).toBeVisible();
+      expect(screen.queryByText("Option 1", { selector })).toBeVisible();
+      expect(screen.queryByText("Option 2", { selector })).toBeVisible();
+      expect(screen.queryByText("Option 3", { selector })).toBeVisible();
+      expect(screen.queryByText("Option 44", { selector })).toBeVisible();
+    });
   });
```
evshi, line 552: Shouldn't this not be visible? Can we make this into a separate test?

PR #1005, opened 2025-01-22
fix: modularize icons

### LINKED ISSUE #185: Add support for flowbite icons

sagardspeed2:
## Describe your changes
reduce runtime bundle size by modularize icons

Closes #185 

- [ ] I have added tests


packages/framework/src/shared/dto.ts
```diff
@@ -63,3 +63,13 @@ export interface FontDTO {
...
   readonly fontStyle: string;
 }
+
+export interface IconSet {
+  readonly prefix?: string;
+  readonly icons?: { [key: string]: unknown };
+}
+
+export interface IconDTO {
+  readonly mui: IconSet;
```
evshi, line 73: There shouldn't be a special case for mui.

packages/runtime/src/EnsembleApp.tsx
```diff
@@ -58,6 +69,13 @@ export const EnsembleApp: React.FC<EnsembleAppProps> = ({
         );
       });
 
+      if (!appDto.icons?.mui) {
+        throw new Error("An mui icons must be provided");
+      }
+
+      registerIcons(appDto.icons.mui);
+      registerIcons(appDto.icons?.custom);
```
evshi, line 77: This should be a warning.

packages/runtime/src/shared/styles.ts
```diff
@@ -97,8 +96,10 @@ export const getCrossAxis = (crossAxis: string): string | undefined => {
   }
 };
 
-export const getIcon = (name: string): SvgIconComponent | undefined => {
-  return get(Icons, name) as SvgIconComponent;
+export const getIcon = <T = React.ComponentType<any>>(
+  name: string,
+): T | undefined => {
+  return IconRegistry.find(name) as T | undefined;
```
evshi, line 102: Don't cast here.

evshi (commented):
This should be generalized to allow any IconSet, not just specially handing MUI Icons.

If you don't import the entire icon set and instead only import the ones that are used, does that reduce the bundle size further?

packages/runtime/src/runtime/menu.tsx
```diff
@@ -425,7 +401,7 @@ const MenuItems: React.FC<{
         {items.map((item, itemIndex) => (
           <AntMenu.Item
             data-testid={item.id ?? item.testId}
-            icon={getIcon(item)}
+            icon={<span>{getCustomIcon(item)}</span>}
```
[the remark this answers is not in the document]
sagardspeed2: In menu, there is space between icon and label, and for that we need here one extra span, which will add space between icon and label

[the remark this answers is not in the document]
evshi: This should be done with CSS instead.

packages/runtime/src/EnsembleApp.tsx
```diff
@@ -58,6 +61,20 @@ export const EnsembleApp: React.FC<EnsembleAppProps> = ({
         );
       });
 
+      Object.entries(Icons).forEach(([name, icon]) => {
+        IconRegistry.register(name, icon as WidgetComponent<unknown>);
+      });
```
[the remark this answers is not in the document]
sagardspeed2: I checked by only importing couple of Icons from MUI, and it did not reduce any bundle size

packages/runtime/src/widgets/Icon.tsx
```diff
@@ -50,36 +50,39 @@ export const Icon: React.FC<IconProps> = ({
...
+      borderStyle: values?.styles?.borderWidth ? "solid" : undefined,
+      ...(values?.styles?.visible === false ? { display: "none" } : undefined),
+    },
+  };
+
+  if (isValidElement(IconComponent)) {
+    return React.cloneElement(IconComponent, { ...iconProps });
+  }
+
+  return <IconComponent {...iconProps} />;
```
evshi, line 85: What is the difference between these two lines?

packages/runtime/src/widgets/Avatar/Avatar.tsx
```diff
@@ -108,20 +110,24 @@ export const Avatar: React.FC<AvatarProps> = (props) => {
...
-                  />
-                </ListItemIcon>
-              ) : null}
-              {menuItem.label}
-            </MenuItem>
-          ))}
+          {values?.menu.map((menuItem, idx) => {
+            const icon = isString(menuItem.icon)
+              ? { name: menuItem.icon }
+              : menuItem.icon;
```
evshi, line 116: This can be a common util to convert either a string or object props to an icon component.

packages/runtime/src/widgets/Icon.tsx
sagardspeed2: `React.cloneElement()` is for modifying an existing React element's props, while `<IconComponent />` is for creating a new element from a component. That's why we need that check - to handle both scenarios since our Icon can be used either as a basic icon (like in Collapsible expand/collapse) or as a full widget in runtime rendering.

evshi: Something is not right here - we shouldn't be getting elements in here. Everything should be a component.

sagardspeed2: You can check `Menu` and `Collapsible` widgets where we are using `icon` as element, and else where like yaml we are using that as `widget`

evshi: If they are being used as widgets in YAML then they would be rendered through Ensemble.runtime

evshi (approved):

evshi (changes requested):
Can you add some tests for Icon/Avatar component.

PR #1008, opened 2025-01-23
added support of dynamic searchKey in search widget

### LINKED ISSUE #1006: Can't pass `searchKey` a dynamic value
**Describe the bug**
We can only pass string in searchKey, but  need to pass dynamic value to make it reusable, i.e like this `searchKey: ${searchKey}`
if we pass it like this, searching working, but it doesn't select the value

getting value like this: `value: "undefined_2"`
Image

sagardspeed2:

Closes #1006 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Search.tsx
+++ b/packages/runtime/src/widgets/Search.tsx
@@ -65,7 +65,7 @@ export const Search: React.FC<SearchProps> = ({
   });
 
   const { id, rootRef, values } = useRegisterBindings(
-    { styles, value, ...rest, widgetName, initialValue },
+    { styles, value, ...rest, widgetName, initialValue, searchKey },
     rest.id,
     {
       setValue,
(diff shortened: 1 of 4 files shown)


evshi (changes requested):
Needs test

packages/runtime/src/widgets/__tests__/Search.test.tsx
```diff
@@ -0,0 +1,56 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import "@testing-library/jest-dom";
+import { BrowserRouter } from "react-router-dom";
+import userEvent from "@testing-library/user-event";
+import { EnsembleScreen } from "../../runtime/screen";
+import "../index";
+
+describe("Search Widget", () => {
+  test("searchKey test", async () => {
```
evshi, line 9: rename this to dynamic search key.

evshi (commented):
can you add a test for a static search key also.

packages/runtime/src/widgets/Search.tsx
```diff
@@ -83,7 +83,7 @@ export const Search: React.FC<SearchProps> = ({
         option,
         values?.searchKey
           ? [itemTemplate?.name ?? "", values.searchKey]
-          : [(itemTemplate?.value || itemTemplate?.name) ?? ""],
+          : itemTemplate?.value || itemTemplate?.name || "",
```
evshi, line 86: Did you mean to remove this here?

evshi (approved):

PR #1012, opened 2025-01-27
fix: added useCommandCallback in uploadFiles

sagardspeed2:
## Describe your changes
Added useCommandCallback in uploadFiles

Closes #

- [x] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
+++ b/packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
@@ -48,6 +48,7 @@ import {
   keys,
   last,
   toNumber,
+  noop,
 } from "lodash-es";
 import { useState, useEffect, useMemo, useCallback, useContext } from "react";
 import { useLocation, useNavigate } from "react-router-dom";
@@ -562,7 +563,6 @@ export const useUploadFiles: EnsembleActionHook<UploadFilesAction> = (
 ) => {
   const screenModel = useScreenModel();
   const appContext = useApplicationContext();
-  const navigate = useNavigate();
   const [body, setBody] = useState<{ [key: string]: unknown }>();
   const [headers, setHeaders] = useState<{ [key: string]: unknown }>();
   const [status, setStatus] = useState<UploadStatus>("pending");
@@ -654,7 +654,7 @@ export const useUploadFiles: EnsembleActionHook<UploadFilesAction> = (
         onErrorAction?.callback({ error });
       }
     },
-    { navigate },
+    { navigate: noop },
     [action, apiModel, screenModel, appContext],
   );
 
--- a/packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
+++ b/packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
@@ -24,7 +24,6 @@ import type {
   EnsembleAction,
   PickFilesAction,
   UploadFilesAction,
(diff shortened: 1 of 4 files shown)


packages/runtime/src/widgets/__tests__/Button.test.tsx
```diff
@@ -537,6 +556,72 @@ describe("Button Widget File Upload Tests", () => {
...
+      userEvent.upload(pickFiles, files);
+    });
+
+    // First verify file is uploaded
+    await waitFor(() => {
+      expect(pickFiles.files).toHaveLength(1);
+      expect(pickFiles.files?.[0].name).toBe("example.png");
+    });
+
+    jest.advanceTimersByTime(5000);
```
evshi, line 619: You shouldn't need this? `waitFor` already has a timeout built in.

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -561,6 +560,9 @@ export const usePickFiles: EnsembleActionHook<PickFilesAction> = (
 export const useUploadFiles: EnsembleActionHook<UploadFilesAction> = (
   action?: UploadFilesAction,
 ) => {
+  const screenModel = useScreenModel();
+  const appContext = useApplicationContext();
+  const navigate = useNavigate();
```
evshi, line 565: I don't think you need this hook you can just pass in `noop`.

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -600,54 +602,60 @@ export const useUploadFiles: EnsembleActionHook<UploadFilesAction> = (
...
   }, []);
 
-  const evaluatedInputs = useEvaluate(action?.inputs);
+  const callback = useCommandCallback(
+    async (evalContext, ...args) => {
+      if (!action) return;
+
+      if (!action.uploadApi || !apiModel) return;
+
+      const context = merge({}, evalContext, args[0]);
```
evshi, line 611: Isn't this built into useCommandCallback already?

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
sagardspeed2: You mean pass this all as null ?

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
sagardspeed2: No, we need to merge the evalContext of useCommandCallback with the args that has been passed with actions. This is not implemented in useCommandCallback and we are doing this in all actions

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
evshi: Jut navigate API - there's a lodash util called `noop` that is an empty function.

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
evshi: Hmmm okay we can refactor this later.

evshi:
Can you also move the test out of Button.test.tsx and into useUploadFiles.test.tsx

evshi (approved):

PR #1016, opened 2025-01-29
enhance MultiSelect component with maxCount, maxTagCount, and maxTagTextLength props

anserwaseem:

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Form/MultiSelect.tsx
+++ b/packages/runtime/src/widgets/Form/MultiSelect.tsx
@@ -14,7 +14,15 @@ import {
 } from "@ensembleui/react-framework";
 import { PlusCircleOutlined } from "@ant-design/icons";
 import { Select as SelectComponent, Space, Form } from "antd";
-import { get, isArray, isEmpty, isEqual, isObject, isString } from "lodash-es";
+import {
+  get,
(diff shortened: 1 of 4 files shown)


packages/runtime/src/widgets/Form/MultiSelect.tsx
```diff
@@ -59,6 +67,12 @@ export type MultiSelectProps = {
   } & EnsembleAction;
   hintStyle?: EnsembleWidgetStyles;
   allowCreateOptions?: boolean;
+  /**	The max number of items can be selected */
+  maxCount: Expression<number>;
+  /** Max tag count to show */
+  maxTagCount: Expression<number | "responsive">;
```
evshi, line 73: What are the differences between this two?

evshi (commented):
Can you add a couple tests?

packages/runtime/src/widgets/Form/MultiSelect.tsx
anserwaseem: maxTagCount controls the visibility of tags after which `+ ${number} ...` will be indicated
maxCount controls the max number of allowed selections

evshi (approved):

PR #1018, opened 2025-02-03
1014-memoize-conditional-branch-widgets

### LINKED ISSUE #1014: Improve conditional performance by memoizing branch widgets
Currently conditional widget lazily renders the widget when the condition is met.

Memoize the widgets in each branch so they don't need to be re-rendered by React.

**Additional context**
This should be helpful when conditionally rendering large widgets like DataGrids.

sagardspeed2:
## Describe your changes
Memoize conditional branch widgets

Closes #1014 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Conditional.tsx
+++ b/packages/runtime/src/widgets/Conditional.tsx
@@ -1,7 +1,7 @@
-import type { Expression } from "@ensembleui/react-framework";
+import type { EnsembleWidget, Expression } from "@ensembleui/react-framework";
 import { unwrapWidget, useRegisterBindings } from "@ensembleui/react-framework";
 import { cloneDeep, head, isEmpty, last } from "lodash-es";
-import { useMemo } from "react";
+import { useMemo, useState } from "react";
 import { WidgetRegistry } from "../registry";
 import { EnsembleRuntime } from "../runtime";
 import type { EnsembleWidgetProps } from "../shared/types";
@@ -26,6 +26,7 @@ export const Conditional: React.FC<ConditionalProps> = ({
   conditions,
   ...props
 }) => {
+  const [matched, setMatched] = useState<{ [key: string]: unknown }>({});
   const [isValid, errorMessage] = hasProperStructure(conditions);
   if (!isValid) throw Error(errorMessage);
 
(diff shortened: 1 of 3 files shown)


packages/runtime/src/widgets/Conditional.tsx
```diff
@@ -26,6 +27,7 @@ export const Conditional: React.FC<ConditionalProps> = ({
   conditions,
   ...props
 }) => {
+  const [matched, setMatched] = useState<{ [key: string]: unknown }>({});
```
evshi, line 30: This should be a ref, otherwise this is unnecessary rerenders.

packages/runtime/src/widgets/Conditional.tsx
```diff
@@ -55,18 +57,27 @@ export const Conditional: React.FC<ConditionalProps> = ({
     if (trueIndex === undefined || trueIndex < 0) {
       return null;
     }
+    const key = conditionStatements[trueIndex]?.toString();
+
+    if (key && matched[key]) {
+      return matched[key] as ReactNode[];
```
evshi, line 63: This cast is not necessary, the type should be in the map.

packages/runtime/src/widgets/Conditional.tsx
```diff
@@ -55,18 +57,27 @@ export const Conditional: React.FC<ConditionalProps> = ({
...
+    }
+    return renderWidget;
   }, [conditionStatements, conditions, trueIndex]);
 
   if (!widget) {
     return null;
   }
 
-  return <>{EnsembleRuntime.render([widget])}</>;
+  return <>{widget}</>;
```
evshi, line 80: Don't need a fragment here anymore.

sagardspeed2: Then we also need to update the return type of the `Conditional` widget from `React.FC`. is this fine ??

evshi: Why would that be the case? `widget` should be a `ReactNode`

sagardspeed2: The `EnsembleRuntime.render` function returns `ReactNode[]`. When we save this in our widget memo, it will contain either that array of nodes or null. so we need to wrap the output in a fragment (<>...</>).

evshi: I'm not sure why this matters since `ReactNode[]` is also a valid `ReactNode`.

type ReactNode =
        | ReactElement
        | string
        | number
        | Iterable<ReactNode>
        | ReactPortal
        | boolean
        | null
        | undefined
        | DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES[
            keyof DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES
        ];

packages/runtime/src/widgets/__tests__/Conditional.test.tsx
```diff
@@ -233,3 +250,91 @@ describe("extractCondition Function", () => {
...
+    expect(unwrapWidgetSpy).toHaveBeenCalledTimes(3);
+    expect(screen.getByText("Greater than 0")).not.toBeNull();
+
+    fireEvent.click(screen.getByText("Decrease"));
+    expect(unwrapWidgetSpy).toHaveBeenCalledTimes(3);
+    expect(screen.getByText("Equals to 0")).not.toBeNull();
+
+    fireEvent.click(screen.getByText("Decrease"));
+    expect(unwrapWidgetSpy).toHaveBeenCalledTimes(3);
+    expect(screen.getByText("Less than 0")).not.toBeNull();
```
evshi, line 338: Can you add assertion about how many times EnsembleRuntime.render was called?

PR #1024, opened 2025-02-05
fix: added support for bindings in invokeAPI

### LINKED ISSUE #848: Can't use bindings in invoke API name
**Describe the bug**
Using a binding in the invokeAPI.name property causes React to throw an error about hooks order because of we use hooks to handle the APIs onResponse and onError. On initial render, the API model isn't resolved yet so the hooks are not registered. Once the API model is resolved, then the hooks are called and that creates the error.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

sagardspeed2:
## Describe your changes
Added support for bindings in invokeAPI

Closes #848 

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/api/data.ts
+++ b/packages/framework/src/api/data.ts
@@ -11,6 +11,7 @@ import type {
 import { screenDataAtom, type ScreenContextDefinition } from "../state";
 import { isUsingMockResponse } from "../appConfig";
 import { mockResponse } from "../evaluate/mock";
+import { evaluateDeep } from "../evaluate";
 
 export const invokeAPI = async (
   apiName: string,
@@ -21,8 +22,14 @@ export const invokeAPI = async (
   setter?: Setter,
   options?: InvokeAPIOptions,
 ): Promise<Response | undefined> => {
+  const evaluated = evaluateDeep(
+    { apiName },
+    screenContext.model,
+    screenContext,
+  );
+
   const api = screenContext.model?.apis?.find(
-    (model) => model.name === apiName,
+    (model) => model.name === evaluated.apiName,
   );
 
   const hash = generateAPIHash({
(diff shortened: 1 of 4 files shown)


packages/framework/src/api/data.ts
```diff
@@ -21,8 +22,14 @@ export const invokeAPI = async (
   setter?: Setter,
   options?: InvokeAPIOptions,
 ): Promise<Response | undefined> => {
+  const evaluated = evaluateDeep(
```
evshi, line 25: How does this work with the `useInvokeAPI` hook? That hook clearly checks if the API name exists before invoking.

packages/runtime/src/runtime/hooks/__tests__/useInvokeApi.test.tsx
```diff
@@ -444,3 +444,43 @@ test("after API fetching using toggle check states", async () => {
...
+          executeCode: "ensemble.storage.set('apiName', 'fetchFoobar')",
+        },
+      }}
+    />,
+    { wrapper: BrowserRouterWrapper },
+  );
+
+  fireEvent.click(screen.getByText("Call API"));
+
+  expect(fetchMock).toHaveBeenCalledTimes(1);
```
evshi, line 485: Add specific params for what fetchMock is being called with.

packages/framework/src/api/data.ts
sagardspeed2: if you check `useEnsembleAction`, in `useInvokeAPI` we are using `useRegisterBindings` for evaluate API name and then we check that is API name exists in `screenContext` before invoking the api, and here when user trigger the api through `ensemble.invokeAPI`, we are using `evaluateDeep` to evaluate the api name and then we are checking that name with `screenContext` and if we found the api, only then we invoke it

PR #1030, opened 2025-02-07
fix: add global or import scripts into DOM

### LINKED ISSUE #1027: Improve performance with large global scripts
When we have large global scripts on a screen, binding performance is impacted because each binding must parse through the global script.

Add the global script to the document so individual bindings do not have to reparse

**Additional context**
See how much of the Text widget render time is due to compiling the JS
Image

sagardspeed2:
## Describe your changes
Add global or import scripts into DOM to reduce the load on buildEvaluateFunction

Closes #1027 

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/evaluate/evaluate.ts
+++ b/packages/framework/src/evaluate/evaluate.ts
@@ -38,14 +38,9 @@ export const buildEvaluateFn = (
       // Need to filter out invalid JS identifiers
     ].filter(([key, _]) => !key.includes(".")),
   );
-  const globalBlock = screen.model?.global;
-  const importedScriptBlock = screen.model?.importedScripts;
 
   // eslint-disable-next-line @typescript-eslint/no-implied-eval, no-new-func
-  const jsFunc = new Function(
-    ...Object.keys(invokableObj),
-    addScriptBlock(formatJs(js), globalBlock, importedScriptBlock),
-  );
+  const jsFunc = new Function(...Object.keys(invokableObj), formatJs(js));
 
   // eslint-disable-next-line @typescript-eslint/no-unsafe-return
   return () => jsFunc(...Object.values(invokableObj));
@@ -78,24 +73,6 @@ const formatJs = (js?: string): string => {
   }
 
   return `return ${sanitizedJs}`;
-};
-
-const addScriptBlock = (
-  js: string,
-  globalBlock?: string,
-  importedScriptBlock?: string,
-): string => {
-  let jsString = ``;
-
-  if (importedScriptBlock) {
-    jsString += `${importedScriptBlock}\n\n`;
-  }
-
-  if (globalBlock) {
-    jsString += `${globalBlock}\n\n`;
-  }
-
-  return (jsString += `${js}`);
 };
 
 /**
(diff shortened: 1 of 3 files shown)


packages/runtime/src/runtime/screen.tsx
```diff
@@ -91,6 +91,35 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
...
 
+  useEffect(() => {
+    const globalBlock = screen.global;
+    const importedScripts = screen.importedScripts;
+
+    const isScriptExist = document.getElementById("custom-scope-script");
+
+    const jsString = `
+      ${importedScripts || ""}
+      ${globalBlock || ""}
```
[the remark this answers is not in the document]
evshi: You will still need to define ensemble in the window.

packages/runtime/src/runtime/screen.tsx
```diff
@@ -91,6 +91,52 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
     };
   }, [screen.customWidgets]);
 
+  useEffect(() => {
+    const globalBlock = screen.global;
+    const importedScripts = screen.importedScripts;
+
+    const isScriptExist = document.getElementById("custom-scope-script");
```
evshi, line 98: This needs to be unique per screen.

packages/runtime/src/runtime/screen.tsx
```diff
@@ -91,6 +91,52 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
...
+            return eval('(' + scriptToExecute.toString() + ')()');
+          }
+        }
+      }
+      
+      const evalInClosure = createEvalClosure()
+    `;
+
+    if (isScriptExist) {
+      isScriptExist.textContent = jsString;
```
evshi, line 129: This does not rerun the script, you need to remove the node and reinsert it.

packages/framework/src/evaluate/evaluate.ts
```diff
@@ -38,14 +50,27 @@ export const buildEvaluateFn = (
...
+
+  if (has(invokableObj, "ensemble")) {
+    const tempEnsemble = get(invokableObj, "ensemble") as {
+      [key: string]: unknown;
+    };
+    (window as unknown as InvokableWindow).ensemble = omitBy(
+      tempEnsemble,
+      isUndefined,
+    );
+  }
```
evshi, line 62: This can be set once - most of the functions in ensemble API do not require any screen params. The only thing would be the modal context, which can be passed in as an argument.

packages/runtime/src/runtime/screen.tsx
```diff
@@ -91,6 +91,52 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
...
+        enumerable: true
+      });
+
+      const createEvalClosure = () => {
+        ${importedScripts || ""}
+        ${globalBlock || ""}
+        
+        return (scriptToExecute, context) => {
+          with (context) {
+            return eval('(' + scriptToExecute.toString() + ')()');
```
evshi, line 120: This can just be `scriptToExecute()`

PR #1037, opened 2025-02-12
fix: ensemble preview handle

sagardspeed2:
## Describe your changes
fix: ensemble preview breaking change

Closes #

- [ ] I have added tests

### DIFF
--- a/apps/preview/src/AppPreview.tsx
+++ b/apps/preview/src/AppPreview.tsx
@@ -125,7 +125,7 @@ export const AppPreview: React.FC<{ db: Firestore }> = ({ db }) => {
           bypassCache: boolean;
           screen?: {
             id: string;
-            content: string;
+            editorContent: string;
           };
         };
         if (msg.type === "reload") {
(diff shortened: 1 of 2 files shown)


apps/preview/src/AppPreview.tsx
```diff
@@ -134,7 +134,7 @@ export const AppPreview: React.FC<{ db: Firestore }> = ({ db }) => {
           if (msg.screen) {
             localStorage.setItem(
               `flutter.${msg.screen.id}`,
-              msg.screen.content,
+              JSON.stringify(msg.screen.editorContent),
```
evshi, line 137: Did you test this? I believe this also needs the app id before the screen id but perhaps the react preview is still separate.

sagardspeed2: Yes, I tested that with ensemble studio on local server, and yes react preview is separate

sagardspeed2: there was some payload change in ensemble web studio 2 months ago, so based on that we update the handler in Preview

evshi (approved):

PR #1040, opened 2025-02-13
feat: add option to allow clearing selection in dropdown

### LINKED ISSUE #1039: Allow Clear Option for Dropdown widget
Just like `MultiSelect` has a way of allowing user to clear the selected value by using the cross icon, the Dropdown should also has functionality.

MultiSelect:

Image

**Additional context**

justEhmadSaeed:

Closes #1039

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Form/Dropdown.tsx
+++ b/packages/runtime/src/widgets/Form/Dropdown.tsx
@@ -303,7 +303,7 @@ const Dropdown: React.FC<DropdownProps> = (props) => {
       <div ref={rootRef} style={{ flex: 1, ...formItemStyles }}>
         <EnsembleFormItem values={values}>
           <Select
-            allowClear={values?.allowClear || false}
+            allowClear={values?.allowClear ?? true}
             className={`${values?.styles?.names || ""} ${id}_input`}
             defaultValue={values?.value}
             disabled={values?.enabled === false}
--- a/packages/runtime/src/widgets/Form/Dropdown.tsx
+++ b/packages/runtime/src/widgets/Form/Dropdown.tsx
@@ -48,6 +48,7 @@ export interface SelectOption {
 }
 
 export type DropdownProps = {
+  allowClear?: boolean;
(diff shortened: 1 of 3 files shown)


packages/runtime/src/widgets/Form/Dropdown.tsx
```diff
@@ -302,6 +303,7 @@ const Dropdown: React.FC<DropdownProps> = (props) => {
       <div ref={rootRef} style={{ flex: 1, ...formItemStyles }}>
         <EnsembleFormItem values={values}>
           <Select
+            allowClear={values?.allowClear || false}
```
evshi, line 306: Does flutter have this behavior by default? Maybe it should be true by default?

evshi, line 306: Nit: could just use double bang !!

evshi (approved):

packages/runtime/src/widgets/Form/Dropdown.tsx
justEhmadSaeed: @evshi eslint doesn't allow me to do that. It suggests using Boolean as its alternative

evshi: Then listen to eslint :)

anserwaseem, line 306: can we keep it to true by default conforming with behavior of other such widgets like MultiSelect and Search

justEhmadSaeed: okay

justEhmadSaeed: @evshi had to change it to use `??` to make it true by default

justEhmadSaeed: @anserwaseem  updated

anserwaseem (approved):

PR #1045, opened 2025-02-19
add onKeyDown action in TextInput widget

anserwaseem:

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/Form/TextInput.tsx
+++ b/packages/runtime/src/widgets/Form/TextInput.tsx
@@ -5,7 +5,7 @@ import { useEffect, useMemo, useState, useCallback, useRef } from "react";
 import type { RefCallback, FormEvent } from "react";
 import { runes } from "runes2";
 import type { Rule } from "antd/es/form";
-import { forEach } from "lodash-es";
+import { forEach, isObject, omitBy } from "lodash-es";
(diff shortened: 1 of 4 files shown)


evshi (changes requested):
This should be implemented as a generic key handler and then call form.submit. This is a very non standard interaction. Normally, you would need to at least tab to the submit button to submit the form.

packages/runtime/src/widgets/Form/TextInput.tsx
```diff
@@ -280,8 +286,9 @@ export const TextInput: React.FC<TextInputProps> = (props) => {
             disabled={values?.enabled === false}
             onChange={(event): void => handleChange(event.target.value)}
             {...(values?.inputType === "number" && {
-              onInput: (event): void => handleKeyDown(event),
+              onInput: (event): void => sanitizeNumberInput(event),
```
evshi, line 289: Does this work the way you would expect with a number input? Can we add a test for this?

anserwaseem: i just renamed the method, didn't add any new implementation

anserwaseem: added test as well

evshi (approved):

PR #1052, opened 2025-02-27
Added support for action callbacks in ensemble.invokeAPI

sagardspeed2:

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/hooks/useScreenData.ts
+++ b/packages/framework/src/hooks/useScreenData.ts
@@ -1,6 +1,7 @@
 import { useAtom, useAtomValue } from "jotai";
 import { useCallback, useMemo } from "react";
 import isEqual from "react-fast-compare";
+import { clone } from "lodash-es";
 import type { Response, WebSocketConnection } from "../data";
 import type {
   EnsembleAPIModel,
@@ -19,13 +20,14 @@ export const useScreenData = (): {
     name: string,
     response: Partial<Response> | WebSocketConnection,
   ) => void;
+  setApi: (apiData: EnsembleAPIModel) => void;
   mockResponses: {
     [apiName: string]: EnsembleMockResponse | string | undefined;
   };
 } => {
-  const apis = useAtomValue(screenApiAtom);
   const sockets = useAtomValue(screenSocketAtom);
   const [data, setDataAtom] = useAtom(screenDataAtom);
+  const [apis, setApiAtom] = useAtom(screenApiAtom);
 
   const apiMockResponses = useMemo(() => {
     return apis?.reduce(
@@ -53,11 +55,24 @@ export const useScreenData = (): {
     [data, setDataAtom],
   );
 
+  const setApi = useCallback(
+    (apiData: EnsembleAPIModel) => {
+      const index = apis?.findIndex((api) => api.name === apiData.name);
+      if (index === undefined || isEqual(apis?.[index], apiData) || !apis) {
+        return;
+      }
+      apis[index] = apiData;
+      setApiAtom(clone(apis));
+    },
+    [apis, setApiAtom],
+  );
+
   return {
     apis,
     sockets,
     data,
     setData,
+    setApi,
     mockResponses,
   };
 };
--- a/packages/framework/src/state/screen.ts
+++ b/packages/framework/src/state/screen.ts
@@ -3,7 +3,11 @@ import { focusAtom } from "jotai-optics";
 import { assign } from "lodash-es";
 import { atomFamily } from "jotai/utils";
 import { type Response, type WebSocketConnection } from "../data";
-import type { EnsembleAppModel, EnsembleScreenModel } from "../shared";
+import type {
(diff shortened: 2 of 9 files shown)


packages/runtime/src/runtime/screen/index.tsx
```diff
@@ -102,6 +103,7 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
       screen={screen}
     >
       <ModalWrapper>
+        <ScreenApiWrapper />
```
evshi, line 106: This needs to block the rest of the tree from rendering until all the actions are evaluated properly otherwise onLoad might not invoke the API callback properly.

packages/runtime/src/runtime/screen/wrapper.tsx
```diff
@@ -0,0 +1,34 @@
...
+}: {
+  api: EnsembleAPIModel;
+  setApi: (apiData: EnsembleAPIModel) => void;
+}): null => {
+  const onResponseAction = useEnsembleAction(api.onResponse);
+  const onErrorAction = useEnsembleAction(api.onError);
+
+  useEffect(() => {
+    setApi({ ...api, onResponseAction, onErrorAction });
+  }, [api.name]);
```
evshi, line 31: You don't need to set the API again you can just assign it - it should be the same object throughout.

evshi (changes requested):

packages/runtime/src/runtime/screen/wrapper.tsx
```diff
@@ -0,0 +1,34 @@
+import type { EnsembleAPIModel } from "@ensembleui/react-framework";
+import { useScreenData } from "@ensembleui/react-framework";
+import { useEffect } from "react";
+// eslint-disable-next-line import/no-cycle
+import { useEnsembleAction } from "../hooks";
```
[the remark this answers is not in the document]
sagardspeed2: We are using `useEnsembleAction` for processing callbacks of APIs in wrapper through screen and this screen also used in `useNavigateModalScreen`. so it cause the cycle.

packages/framework/src/hooks/useScreenData.ts
```diff
@@ -19,13 +20,14 @@ export const useScreenData = (): {
     name: string,
     response: Partial<Response> | WebSocketConnection,
   ) => void;
+  setApi: (apiData: EnsembleAPIModel) => void;
```
evshi, line 23: This is no longer used.

packages/framework/src/state/screen.ts
```diff
@@ -57,10 +61,17 @@ export const screenModelAtom = focusAtom(screenAtom, (optic) =>
   optic.prop("model"),
 );
 
-export const screenApiAtom = focusAtom(screenAtom, (optic) => {
+export const screenApiFocusAtom = focusAtom(screenAtom, (optic) => {
   return optic.prop("model").optional().prop("apis");
 });
 
+export const screenApiAtom = atom(
```
evshi, line 68: Not needed anymore?

packages/runtime/src/runtime/screen/index.tsx
```diff
@@ -49,8 +51,23 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
...
+      return;
+    }
+
+    screen.apis = processedAPIs;
+    setIsAPIProcessed(true);
+  }, [processedAPIs]);
+
+  useEffect(() => {
+    // Ensure customWidgets is defined before using it
+    const customWidgets = screen?.customWidgets || [];
```
evshi, line 68: This is not needed isEmpty checks for nullish.

packages/runtime/src/runtime/screen/index.tsx
```diff
@@ -49,8 +51,23 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
     inputs,
   ) as { [key: string]: unknown };
 
+  const processedAPIs = processApiDefinitions(screen.apis);
```
evshi, line 54: This should be in useEffect? This should only happen once per render. Make this a hook.

packages/runtime/src/runtime/screen/index.tsx
```diff
@@ -49,8 +47,23 @@ export const EnsembleScreen: React.FC<EnsembleScreenProps> = ({
...
   useEffect(() => {
-    if (!screen.customWidgets || isEmpty(screen.customWidgets)) {
+    if (isEmpty(screen.apis)) {
+      setIsAPIProcessed(true);
+      return;
+    }
+
+    screen.apis = processedAPIs;
+    setIsAPIProcessed(true);
+  }, [processedAPIs, screen]);
```
evshi, line 60: This can all just be one hook.

evshi (commented):
Hooks can do a lot more so you can encapsulate everything into a single hook. Take `useQuery` as an example.

evshi (approved):

PR #1053, opened 2025-02-28
fix: added support for await in executeCode

sagardspeed2:

Closes #

- [x] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/hooks/__tests__/useExecuteCode.test.tsx
+++ b/packages/runtime/src/runtime/hooks/__tests__/useExecuteCode.test.tsx
@@ -101,7 +101,8 @@ test("call ensemble.invokeAPI", async () => {
   const { result } = renderHook(
     () =>
       useExecuteCode(
-        "await ensemble.invokeAPI('getDummyProductsByPaginate', apiConfig)",
+        `const res = await ensemble.invokeAPI('getDummyProductsByPaginate', apiConfig);
+        return res.body.limit`,
         { context: { apiConfig } },
       ),
     {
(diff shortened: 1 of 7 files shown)


packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
```diff
@@ -130,14 +130,28 @@ export const useExecuteCode: EnsembleActionHook<
...
+
+      if (js.includes("\n")) {
+        executableJS = `return (async () => {
+          ${js}
+        })()`;
+      }
+
+      const retVal = executableJS.includes("await ")
+        ? await evaluate({ model: screenModel }, executableJS, context)
+        : evaluate({ model: screenModel }, executableJS, context);
```
evshi, line 153: This needs to be done in `evaluate`.

evshi (changes requested):

packages/runtime/src/runtime/hooks/useEnsembleAction.tsx
sagardspeed2: Hey for that we need to make a big change, because every place where we used `evaluate` we convert that with await as buildEvaluateFn will always return Promise and to fulfill that we need to make `evaluate` as await

evshi (commented):
Can we remove all the async/awaits from the tests that don't need them. And then add a test that actually does use await in the code.

packages/runtime/src/runtime/hooks/__tests__/useExecuteCode.test.tsx
```diff
@@ -101,7 +101,7 @@ test("call ensemble.invokeAPI", async () => {
   const { result } = renderHook(
     () =>
       useExecuteCode(
-        "ensemble.invokeAPI('getDummyProductsByPaginate', apiConfig).then((res) => res.body.products.length)",
+        "await ensemble.invokeAPI('getDummyProductsByPaginate', apiConfig)",
```
sagardspeed2, line 104: Here this test actually demonstrate the use of `await` in js ( useEnsembleAction through evaluate )

evshi:
@sagardspeed2 did you test these changes with the Atlas app? If it is always returning a promise then that may create issues when combined with other actions. Add a unit test for this case.

evshi (approved):

anserwaseem (approved):

PR #1055, opened 2025-03-03
fix: added support for disable row selection

sagardspeed2:
## Describe your changes
Added support for disable row selection

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
+++ b/packages/runtime/src/widgets/DataGrid/DataGrid.tsx
@@ -78,6 +78,7 @@ export interface DataGridStyles extends Partial<EnsembleWidgetStyles> {
 export interface DataGridRowTemplate {
   name: "DataRow";
   properties: {
+    selectable?: Expression<boolean>;
     onTap?: EnsembleAction;
     children?: EnsembleWidget[];
   } & HasItemTemplate;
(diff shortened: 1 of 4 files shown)


packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -449,6 +450,26 @@ export const DataGrid: React.FC<GridProps> = (props) => {
     return [];
   }, [values?.DataColumns, values?.allowResizableColumns, colWidth]);
 
+  const handleRowSelectableOrNot = useCallback(
+    (record: unknown): boolean => {
+      const { selectable } = itemTemplate.template.properties;
+
+      // If selectable is not defined or false, return false immediately
+      if (!selectable) return false;
```
anserwaseem, line 458: let's deal all rows as selectable by default

sagardspeed2: This deal with all rows as selectable, as this props pass to disabled stat, so.

sagardspeed2: `disable: false` will be selectable

anserwaseem: then let's proceed with name `disableSelection`

anserwaseem (approved):

evshi (approved):

PR #1060, opened 2025-03-17
fix: added support for nested items in menu

### LINKED ISSUE #1059: Allow Sidebar's sub menu items to be active when selected
When the fax center or sms center is selected from the communication entry, we should set it to active
Image

**Additional context**

sagardspeed2:
## Describe your changes
Added support for nested items in menu

Closes #1059 

- [ ] I have added tests

### DIFF
--- a/packages/runtime/src/runtime/menu.tsx
+++ b/packages/runtime/src/runtime/menu.tsx
@@ -59,6 +59,7 @@ interface MenuItemProps {
   hasNotifications?: boolean;
   openNewTab?: boolean;
   visible?: boolean;
+  childrens?: MenuItemProps[];
   customItem?: {
     widget?: { [key: string]: unknown };
     selectedWidget?: { [key: string]: unknown };
@@ -165,12 +166,13 @@ export const EnsembleMenu: React.FC<{
   const [isCollapsed, setIsCollapsed] = useState<boolean>(
     type === EnsembleMenuModelType.Drawer,
   );
-
+  const [defaultOpenKeys, setDefaultOpenKeys] = useState<string>();
   const outletContext = {
     isMenuCollapsed: isCollapsed,
     setMenuCollapsed: setIsCollapsed,
   };
   const { id, items: rawItems, styles, header, footer, onCollapse } = menu;
+
   // custom items may contain their own bindings to be evaluated in dynamic context
   const itemInputs = rawItems?.map<MenuItemProps>((item) =>
     omit(item, "customItem"),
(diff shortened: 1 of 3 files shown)


packages/runtime/src/runtime/menu.tsx
```diff
@@ -165,12 +166,13 @@ export const EnsembleMenu: React.FC<{
   const [isCollapsed, setIsCollapsed] = useState<boolean>(
     type === EnsembleMenuModelType.Drawer,
   );
-
+  const [defaultOpenKeys, setDefaultOpenKeys] = useState<string>();
```
justEhmadSaeed, line 169: @sagardspeed2 shouldn't `defaultOpenKeys` be an array as the name suggests and antD is also expecting an array

justEhmadSaeed:
@sagardspeed2 
when we add the children to a menu item, it stops following the sidebar styles.
image

justEhmadSaeed:
also, we might be needing the support for customItem in the sub-menu as we're doing it for main menu items

packages/runtime/src/runtime/menu.tsx
```diff
@@ -422,43 +486,100 @@ const MenuItems: React.FC<{
...
+              key={`submenu-${itemIndex}`}
+              style={{
+                color: styles.labelColor ?? "grey",
+                fontSize: "1rem",
+                height: "auto",
+              }}
+              title={getLabel(item)}
+            >
+              {item.children.map((childItem, childIndex) => (
+                <AntMenu.Item
```
evshi, line 502: This can be put into a function to make it more readable since this is the same as lines 546-580?

packages/runtime/src/runtime/menu.tsx
```diff
@@ -165,12 +167,14 @@ export const EnsembleMenu: React.FC<{
   const [isCollapsed, setIsCollapsed] = useState<boolean>(
     type === EnsembleMenuModelType.Drawer,
   );
-
+  const [defaultOpenKeys, setDefaultOpenKeys] = useState<string[]>([]);
+  const [isInitialized, setIsInitialized] = useState<boolean>(false);
```
evshi, line 171: Why do you need a separate initialized state variable?

sagardspeed2: We need to first check all defaultOpenKeys and the selected key, so we can render this value on the menu widgets.

without this flag, the menu widget is renders with empty defaultOpenKeys and null selected keys and later updating the state of defaultOpenKeys will not toggle the default submenu expand.

also this is useful to prevent unnecessary render until we have all the states updated as per needed

evshi (approved):

PR #1067, opened 2025-04-21
optimize useWidgetId hook by utilizing latest and greatest createBindingAtom

anserwaseem:
## Describe your changes
Problem:
When using any method/value from an imported/global script in `id` of any widget, it throws an error

Closes #

- [ ] I have added tests

### DIFF
--- a/packages/framework/src/hooks/useWidgetId.ts
+++ b/packages/framework/src/hooks/useWidgetId.ts
@@ -1,8 +1,8 @@
 import { useMemo } from "react";
+import { atom, useAtomValue } from "jotai";
 import type { Expression } from "../shared";
-import { error, isExpression } from "../shared";
-import { evaluate } from "../evaluate/evaluate";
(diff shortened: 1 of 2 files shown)


packages/framework/src/hooks/useWidgetId.ts
```diff
@@ -13,31 +13,46 @@ export const useWidgetId = (
...
+    return atom<string | undefined>(id);
+  }, [customScope, id]);
+
+  const resolvedId = useAtomValue(idBindingAtom);
+
+  const resolvedWidgetId = useMemo<string>(() => {
+    const workingId = resolvedId;
+    if (workingId && JS_ID_REGEX.test(String(workingId))) {
+      return String(workingId);
     }
```
evshi, line 34: Let's assert the type of the ID is a string otherwise this would cause errors.

evshi (approved):

PR #1081, opened 2025-06-06
DataRow styles with context

anserwaseem:

Closes #

- [ ] I have added tests


packages/runtime/src/widgets/DataGrid/DataGrid.tsx
```diff
@@ -156,6 +156,41 @@ const ResizableTitle: React.FC<ResizableProps & ResizableState> = (props) => {
   );
 };
 
+const CustomRowWithStyles: React.FC<
```
evshi, line 159: You can just use `memo` around the component definition since a lot of your hooks are dependent on the props.

evshi (approved):

PR #1085, opened 2025-06-26
add the global and imported scripts to the document per screen

anserwaseem:

Closes #

- [ ] I have added tests


anserwaseem:
loaded /maanger/tasks page
image

loaded /workfflow page
image

anserwaseem:
please refer two files each for before & after test results:
[per-binding-test.zip](https://github.com/user-attachments/files/21017809/per-binding-test.zip)

| file | explanation|
|---|---|
| first_load.js | bindings result on first load of /manager/tasks page
| set_page_size_100.js | bindings result on setting page size to 100 on /manager/tasks page

Each entry in is an object that aggregates the timings recorded so far for that specific binding:
| entry | explanation|
|:--:|--|
| count | How many times that binding has been evaluated since the page was loaded.
| total   |  The sum (in ms) of all individual evaluation times.
| avg   |  Mean/average with total / count.
| max  |  The single slowest evaluation observed so far (worst-case latency).
| min   |  The single fastest evaluation observed so far (best-case latency).

